PR #26, opened 2019-06-18
Move to cloudevents v 0.3

anatoliyfedorenko:
Switch to cloudevents 0.3.

### DIFF
--- a/awscodecommit/main.go
+++ b/awscodecommit/main.go
@@ -29,6 +29,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/codecommit"
 	"github.com/aws/aws-sdk-go/service/codecommit/codecommitiface"
 	cloudevents "github.com/cloudevents/sdk-go"
+	"github.com/cloudevents/sdk-go/pkg/cloudevents/types"
 	log "github.com/sirupsen/logrus"
 )
 
@@ -271,6 +272,7 @@ func (clients Clients) sendCommitEvent(commit *codecommit.Commit) error {
 		Context: cloudevents.EventContextV03{
 			Type:            "AWS CodeCommit commit event",
 			Subject:         aws.String("AWS CodeCommit Event"),
+			Source:          *types.ParseURLRef(""),
 			ID:              *commit.CommitId,
 			DataContentType: aws.String("application/json"),
 		}.AsV03(),
@@ -298,6 +300,7 @@ func (clients Clients) sendPREvent(pullRequest *codecommit.PullRequest) error {
 		Context: cloudevents.EventContextV03{
 			Type:            "AWS CodeCommit commit event",
 			Subject:         aws.String("AWS CodeCommit Event"),
+			Source:          *types.ParseURLRef(*pullRequest.AuthorArn),
 			ID:              *pullRequest.PullRequestId,
 			DataContentType: aws.String("application/json"),
 		}.AsV03(),
--- a/awscodecommit/main.go
+++ b/awscodecommit/main.go
@@ -272,7 +272,7 @@ func (clients Clients) sendCommitEvent(commit *codecommit.Commit) error {
 		Context: cloudevents.EventContextV03{
 			Type:            "com.amazon.codecommit.commit",
 			Subject:         aws.String(repoNameEnv + "/" + repoBranchEnv),
-			Source:          *types.ParseURLRef(""),
+			Source:          *types.ParseURLRef("https://git-codecommit." + awsRegionEnv + ".amazonaws.com/v1/repos/" + repoNameEnv),
 			ID:              *commit.CommitId,
 			DataContentType: aws.String("application/json"),
 		}.AsV03(),
@@ -300,7 +300,7 @@ func (clients Clients) sendPREvent(pullRequest *codecommit.PullRequest) error {
 		Context: cloudevents.EventContextV03{
 			Type:            "com.amazon.codecommit.pull_request",
 			Subject:         aws.String(repoNameEnv + "/" + repoBranchEnv),
-			Source:          *types.ParseURLRef(*pullRequest.AuthorArn),
+			Source:          *types.ParseURLRef("https://git-codecommit." + awsRegionEnv + ".amazonaws.com/v1/repos/" + repoNameEnv),
 			ID:              *pullRequest.PullRequestId,
 			DataContentType: aws.String("application/json"),
 		}.AsV03(),
--- a/awscodecommit/main.go
+++ b/awscodecommit/main.go
@@ -103,7 +103,9 @@ func main() {
 		log.Fatal(err)
 	}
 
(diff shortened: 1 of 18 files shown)


awscodecommit/main.go
```diff
@@ -259,17 +265,28 @@ func (clients Clients) preparePullRequests() ([]*codecommit.PullRequest, error)
...
 
-	codecommitEvent := PushMessageEvent{
-		Commit:           commit,
-		CommitRepository: aws.String(repoNameEnv),
-		CommitBranch:     aws.String(repoBranchEnv),
-		EventSource:      aws.String("aws:codecommit"),
-		AwsRegion:        aws.String(awsRegionEnv),
+	event := cloudevents.Event{
+		Context: cloudevents.EventContextV03{
+			Type:            "AWS CodeCommit commit event",
```
duglin: any reason you're not using a reverse DNS name as the spec suggests?

sebgoa: as in: https://github.com/cloudevents/spec/blob/v0.3/spec.md#type ? com.amazon.sqs

we are also missing `specVersion` I guess

anatoliyfedorenko: @sebgoa `specVersion` is automatically added by this function `AsV03()` which is why I did not include it in the event description. 

Regarding reverse DNS name, will add it as well

duglin: do you have an example one of these events from AWS? I tried to find one via google but couldn't. That will help me with a recommendation for what the reverse DNS name should be.

awscodecommit/main.go
```diff
@@ -298,7 +298,7 @@ func (clients Clients) sendPREvent(pullRequest *codecommit.PullRequest) error {
 
 	event := cloudevents.Event{
 		Context: cloudevents.EventContextV03{
-			Type:            "AWS CodeCommit commit event",
+			Type:            "com.amazon.codecommit",
```
duglin: as a prefix this is probably correct. But is there something we should be adding to the end to make it more specific? E.g. `push` or something like that related to the action? If this is a static string then it's useful to know it's from CodeCommit instead of GitHub but it's not useful to know what type of action caused this event.

awscognito/main.go
```diff
@@ -236,7 +236,7 @@ func (clients Clients) sendCognitoEvent(dataset *cognitosync.Dataset, records []
 
 	event := cloudevents.Event{
 		Context: cloudevents.EventContextV03{
-			Type:            "SyncTrigger",
+			Type:            "com.amazon.cognito",
```
duglin: same comment here... is there something we an add to the end?

awskinesis/main.go
```diff
@@ -207,7 +207,7 @@ func (clients Clients) sendCognitoEvent(record *kinesis.Record, shardID *string)
 
 	event := cloudevents.Event{
 		Context: cloudevents.EventContextV03{
-			Type:            "AWS Kinesis Record",
+			Type:            "com.amazon.kinesis",
```
duglin: Do you view Kinesis as the event producer or more like an event transport? If it's more like a transport, then I'm not sure this is the right value since it doesn't tell you anything about the event itself.

awssns/main.go
```diff
@@ -215,7 +215,7 @@ func (clients Clients) HandleNotification(w http.ResponseWriter, r *http.Request
 
 		event := cloudevents.Event{
 			Context: cloudevents.EventContextV03{
-				Type:            data["Type"].(string),
+				Type:            "com.amazon.sns",
```
duglin: same comment here. Isn't SNS more like a transport than something that describes the event?

awssqs/main.go
```diff
@@ -199,7 +199,7 @@ func (clients Clients) sendSQSEvent(msg *sqs.Message, queueARN *string) error {
 
 	event := cloudevents.Event{
 		Context: cloudevents.EventContextV03{
-			Type:            "AWS SQS Message",
+			Type:            "com.amazon.sqs",
```
duglin: Ditto here. I think we need to extract something from the event itself - if possible.

awscodecommit/main.go
```diff
@@ -259,17 +265,28 @@ func (clients Clients) preparePullRequests() ([]*codecommit.PullRequest, error)
...
-	codecommitEvent := PushMessageEvent{
-		Commit:           commit,
-		CommitRepository: aws.String(repoNameEnv),
-		CommitBranch:     aws.String(repoBranchEnv),
-		EventSource:      aws.String("aws:codecommit"),
-		AwsRegion:        aws.String(awsRegionEnv),
+	event := cloudevents.Event{
+		Context: cloudevents.EventContextV03{
+			Type:            "AWS CodeCommit commit event",
+			Subject:         aws.String("AWS CodeCommit Event"),
```
duglin: `subject`, if present, should be more descriptive. It needs to be a unique descriptor (URL) of the entity upon which the event happened. For example, perhaps the repo, or branch, depending on the type of event.

awscodecommit/main.go
anatoliyfedorenko: @duglin , maybe adding actual event type would be OK? like `commit` or `merge` etc?

awscodecommit/main.go
```diff
@@ -259,17 +265,28 @@ func (clients Clients) preparePullRequests() ([]*codecommit.PullRequest, error)
...
+			Source:          *types.ParseURLRef(""),
+			ID:              *commit.CommitId,
```
duglin: Can one commitID generate more than one event? If so, this wouldn't be a good value to use since the ID needs to be unique across all events from the same event source.

awscognito/main.go
```diff
@@ -227,7 +234,19 @@ func (clients Clients) sendCognitoEvent(dataset *cognitosync.Dataset, records []
 		DatasetRecords:   records,
 	}
 
-	if err := clients.CloudEvents.Send(cognitoEvent); err != nil {
+	event := cloudevents.Event{
+		Context: cloudevents.EventContextV03{
+			Type:            "com.amazon.cognito",
+			Subject:         aws.String("AWS Cognito"),
+			Source:          *types.ParseURLRef(""), // need to figure out how to indlude it
```
duglin: s/indlude/include/  :-)

awscodecommit/main.go
duglin: this applies to the rest of the PR - so I won't repeat myself

awscodecommit/main.go
duglin: yep - exactly! using `commit` or `merge` at the end of the reverse DNS name is exactly what CE is expecting there.

awskinesis/main.go
anatoliyfedorenko: I guess it is more like event producer here.

anatoliyfedorenko:
@duglin I have made some changes, if you have sometime, please, see if they are appropriate. Thank you for your review!

Issue #30, opened 2020-03-04
CloudEvents update breaks build


sebgoa:
```
$ go build .
# github.com/triggermesh/knative-lambda-sources/awssqs
./main.go:211:9: assignment mismatch: 2 variables but clients.CloudEvents.Send returns 3 values
```

sebgoa:
See https://github.com/cloudevents/sdk-go/blob/master/cmd/samples/http/sender/main.go#L87

a-edakin:
Issue was solved in #28 packages update

PR #31, opened 2020-03-04
prep Dockerfile for RH certification

sebgoa:


awssqs/main.go
```diff
@@ -208,7 +208,7 @@ func (clients Clients) sendSQSEvent(msg *sqs.Message, queueARN *string) error {
 		Data: sqsEvent,
 	}
 
-	_, err := clients.CloudEvents.Send(context.Background(), event)
+	_, _, err := clients.CloudEvents.Send(context.Background(), event)
```
sebgoa: this solves #30 but feels dirty without proper handling :(

PR #36, opened 2020-03-11
Packages update #28

### LINKED ISSUE #28: Need to update our go packages
while building

```
Step 5/12 : RUN go get github.com/golang/dep/cmd/dep &&     dep ensure -v &&     CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -installsuffix cgo -o /go/bin/sqs
 ---> Running in b22261a6c7d4
Warning: the following project(s) have [[constraint]] stanzas in Gopkg.toml:

  ✗  github.com/triggermesh/sources

However, these projects are not direct dependencies of the current project:
they are not imported in any .go files, nor are they in the 'required' list in
Gopkg.toml. Dep only applies [[constraint]] rules to direct dependencies, so
these rules will have no effect.

Either import/require packages from these projects so that they become direct
dependencies, or convert each [[constraint]] to an [[override]] to enforce rules
on these projects, if they happen to be transitive dependencies.

# Gopkg.lock is out of sync with Gopkg.toml and project imports:
github.com/cloudevents/sdk-go/pkg/cloudevents/types: imported or required, but missing from Gopkg.lock's input-imports

Root project is "sqs"
 1 transitively valid internal packages
 10 external packages imported from 5 projects
(0)   ✓ select (root)

```

a-edakin:
Updated packages in awscodecommit, awscognito, awsdynamodb, awskinesis, awssns, awssqs

closes #28 #30

### DIFF
--- a/awscodecommit/Gopkg.lock
+++ b/awscodecommit/Gopkg.lock
@@ -2,9 +2,10 @@
 
 
 [[projects]]
-  digest = "1:fda153e673adba59571d2474e58e2253ed5113ab09feca1c52e2be9e7ec3ca9d"
+  digest = "1:354d36b7971141353fa858bc5c9f354a43896363d008db20190e21480bef7749"
   name = "github.com/aws/aws-sdk-go"
   packages = [
+    ".",
     "aws",
     "aws/awserr",
     "aws/awsutil",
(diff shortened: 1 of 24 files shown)


sebgoa:
Why did you have to re-write the tests ?

a-edakin:
Yeah, it would be great, but also i would need to refactor env loading refactor in main.go because in some sources it is impossible to run tests with just `go test`

a-edakin:
And i noticed that all containers uses `golang:alpine` in my opinion we should specify image version

sebgoa:
I am going to merge this right away, because we also need to move to go modules and pin cloudevents to v1.0

PR #39, opened 2020-03-19
commenting out SNS to fix build timeout

sebgoa:
@sameersbn quick review here.

To get you on board :) you should be able to see where this build is configured in Google Cloud Build, triggermesh-production project...and see the logs.

The build is timing out for some reason.

We will need to decide whether a mono repo makes sense or whether we need to do individual repos for each source.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -27,8 +27,8 @@ steps:
 #  args: [ 'build', '-t', 'gcr.io/$PROJECT_ID/awss3:latest', '-f' ,'./awss3/Dockerfile', '.' ]
 
 ## Build SNS source
-- name: 'gcr.io/cloud-builders/docker'
-  args: [ 'build', '-t', 'gcr.io/$PROJECT_ID/awssns:latest', '-f' ,'./awssns/Dockerfile', '.' ]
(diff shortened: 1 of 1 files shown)


sebgoa:
Allright, it needs be looked over again at some point but:

* use a bigger machine
* parallelize things a bit

Gets it down to 2min 42s compared to timing out at 10 mins...

cloudbuild.yaml
```diff
@@ -2,33 +2,94 @@
...
+  args:
+  - '-c'
+  - |
+    docker pull gcr.io/$PROJECT_ID/awsdynamodb:latest || exit 0
+- name: 'gcr.io/cloud-builders/docker'
+  entrypoint: 'bash'
+  args:
+  - '-c'
+  - |
+    docker pull gcr.io/$PROJECT_ID/awssqs:latest || exit 0
```
sameersbn: We can parallelize the pull with `waitFor: ['-']` which indicates that this step begins immediately.
ref: https://cloud.google.com/cloud-build/docs/configuring-builds/configure-build-step-order

sebgoa:
ok added that it shaved 10 seconds. I don't think the caching I did is helping much....I will merge it for now

PR #43, opened 2020-03-24
Moved from dep to go modules

### LINKED ISSUE #29: Move to Go module
?

a-edakin:
Updated all sources to go modules 
Updated Dockerfiles and README.md for all updated sources 

closes #29


a-edakin:
> It looks like kinesisproducer is still not migrated.
All other changes sounds good.

Wow, i did'n know that it even exists. I'll update it to 

> I'll create a followup issue though to re-order Dockerfiles. Right now a change in code means downloading all go dependencies again.

Yeah, i noticed that to but decide that it would be to match changes

PR #46, opened 2020-04-01
implements common build infrastructure and improve Dockerfile cache hits

### LINKED ISSUE #45: Dockerfiles: re-arrange to have an early dependency layer
Re-arrange Dockerfiles so that dependencies layers are only re-created when go.sum or go.mod are modified.

Copying both files before the rest of the code and issuing a `go mod donwload` at an early step should work.

sameersbn:
In this change we have restructured the Dockerfiles to improve cache hits and also implemented a common build infratructure.
For example you can build all the sources with `make build` or build all images with `make image`. Similarly to test all sources you can do `make test`.

Alternatively if you only want to test the `awskinesis` source, you can do `make -C awskinesis test`.

Execute `make -C awskinesis help` to get the list of all available targets

Closes #45

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,17 @@
+.git/
+.dockerignore
+**/.gitignore
+**/c.out
+**/coverage.html
+**/Dockerfile
+**/README.md
+**/*_test.go
+**/Makefile
+awscodecommit/awscodecommit
+awscognito/awscognito
+awsdynamodb/awsdynamodb
+awsiot/awsiot
+awskinesis/awskinesis
+awss3/awss3
+awssns/awssns
+awssqs/awssqs
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,9 @@
+SUBDIRS ?= awscodecommit awscognito awsdynamodb awskinesis awssqs
+TARGETS := $(shell awk '{FS = ":";} /^[a-zA-Z0-9._-]+:.*?/ { printf "%s ", $$1 }' scripts/inc.Makefile)
+
+.PHONY: $(SUBDIRS) $(TARGETS)
+
+$(TARGETS): $(SUBDIRS)
+
+$(SUBDIRS):
+	@$(MAKE) --no-print-directory -C $@ $(MAKECMDGOALS)
--- a/awscodecommit/Dockerfile
+++ b/awscodecommit/Dockerfile
@@ -1,24 +1,30 @@
-FROM golang:alpine AS builder
+FROM golang:1.14-alpine AS builder
 
-COPY awscodecommit /go/src/awscodecommit
+ENV GO111MODULE on
+ENV CGO_ENABLED 0
+ENV GOOS linux
+ENV GOARCH amd64
+
 WORKDIR /go/src/awscodecommit
 
 RUN apk update && apk add --no-cache openssh-client ca-certificates
-RUN GO111MODULE=on go mod download && \
-    CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -installsuffix cgo -o /go/bin/awscodecommit
+
+COPY awscodecommit/go.mod awscodecommit/go.sum /go/src/awscodecommit/
+RUN go mod download
+
+COPY awscodecommit/ /go/src/awscodecommit/
+RUN go install -installsuffix cgo ./...
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal
 
-LABEL name="Triggermesh AWS CodeCommit Event Source" \
-      vendor="Triggermesh" \
-      version="v0.1.0" \
-      release="1" \
-      summary="The Triggermesh CodeCommit Source" \
-      description="This is the Triggermesh Knative Event Source for AWS CodeCommit"
+LABEL name "Triggermesh AWS CodeCommit Event Source"
+LABEL vendor "Triggermesh"
+LABEL version "v0.1.0"
+LABEL release "1"
+LABEL summary "The Triggermesh CodeCommit Source"
+LABEL description "This is the Triggermesh Knative Event Source for AWS CodeCommit"
 
-RUN microdnf update
-COPY ./licenses/* /licenses/
-
+COPY licenses/ /licenses/
 COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 COPY --from=builder /go/bin/awscodecommit /go/bin/awscodecommit
 
--- a/scripts/inc.Makefile
+++ b/scripts/inc.Makefile
@@ -1,0 +1,53 @@
+RM         ?= rm
+CP         ?= cp
+MV         ?= mv
+
+GO         ?= go
+GOFMT      ?= gofmt
+GOLINT     ?= golint
+GOTEST     ?= go test
+GOTOOL     ?= go tool
+
+DOCKER     ?= docker
+
+LDFLAGS     =
+
+all: build
+
+help: ## Display this help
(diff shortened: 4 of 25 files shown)


sebgoa:
/lgtm

I will let Antoine cut his teeth on that one

.dockerignore
```diff
@@ -0,0 +1,10 @@
+.git/
+.dockerignore
+**/.gitignore
+**/c.out
+**/coverage.html
+**/Dockerfile
+**/README.md
+**/*_test.go
+**/Makefile
+awscodecommit/awscodecommit
```
a-edakin, line 10: I don't understand why here only `awscodecommit/awscodecommit` built binary may be all other sources binaries should be under this `.dockerignore` too?

awscodecommit/Dockerfile
```diff
@@ -1,25 +1,31 @@
-FROM golang:alpine AS builder
+FROM golang:1.13-alpine AS builder
```
sameersbn, line 1: pining go version to make builds more deterministic

antoineco: Just curious, why not 1.14?

.dockerignore
```diff
@@ -0,0 +1,10 @@
+.git/
+.dockerignore
+**/.gitignore
+**/c.out
```
antoineco, line 4: What is generating these?

Makefile
```diff
@@ -0,0 +1,6 @@
+%::
```
antoineco, line 1: Learned [something](https://www.gnu.org/software/make/manual/html_node/Last-Resort.html) today :)

awscodecommit/Dockerfile
```diff
@@ -1,25 +1,31 @@
...
+
+ENV GO111MODULE on
+ENV CGO_ENABLED 0
+ENV GOOS linux
+ENV GOARCH amd64
```
antoineco, line 6: (nit) I know it's less pretty, but we could minimize the number of metadata layers with
```dockerfile
ENV \
    GO111MODULE=on \
    CGO_ENABLED=0 \
    GOOS=linux \
    GOARCH=amd64
```
I don't feel strongly about it though.

awscodecommit/Dockerfile
```diff
@@ -1,25 +1,31 @@
...
+FROM registry.access.redhat.com/ubi8/ubi-minimal
 
-RUN microdnf update
-COPY ./licenses/* /licenses/
+LABEL name "Triggermesh AWS CodeCommit Event Source"
+LABEL vendor "Triggermesh"
+LABEL version "v0.1.0"
+LABEL release "1"
+LABEL summary "The Triggermesh CodeCommit Source"
+LABEL description "This is the Triggermesh Knative Event Source for AWS CodeCommit"
```
antoineco, line 25: Same comment as above, regarding the grouping of `ENV`.

awscodecommit/Dockerfile
```diff
@@ -1,25 +1,31 @@
...
+RUN go mod download
 
-LABEL name="Triggermesh AWS CodeCommit Event Source" \
-      vendor="Triggermesh" \
-      version="v0.1.0" \
-      release="1" \
-      summary="The Triggermesh CodeCommit Source" \
-      description="This is the Triggermesh Knative Event Source for AWS CodeCommit"
+COPY awscodecommit/ /go/src/awscodecommit/
+RUN go install -installsuffix cgo ./...
```
antoineco, line 16: Besides readability, is there any reason not to group the `COPY` statements together and then performing a single `RUN` like in most official Docker images?

a-edakin: [issue 45](https://github.com/triggermesh/knative-lambda-sources/issues/45)

It is made for build optimization, so when we change code in `main.go` we wont have to wait to install unchanged dependencies 

i guess it is made mostly for local development because on CI it wont change anything 

@odacremolbap am i right ?

antoineco: If it's for build speed that makes sense indeed.

scripts/inc.Makefile
```diff
@@ -0,0 +1,53 @@
...
+GOTOOL     ?= go tool
+
+DOCKER     ?= docker
+
+LDFLAGS     =
+
+all: build
+
+help: ## Display this help
+	@awk 'BEGIN {FS = ":.*?## "; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z0-9._-]+:.*?## / {printf "  \033[36m%-20s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
```
antoineco, line 18: Not sure if I understand how this one-liner works, but I just ran `make help` locally and wow, nice!

awscodecommit/Dockerfile
sebgoa: one quick comment about those labels so that everyone knows. They are mandatory to pass the RedHat certification scanning. You can edit them but don't delete them

awscodecommit/Dockerfile
a-edakin: Didn't know that 1.14 was released when i was updating sources 

I think nothing bad will happen if we change version to 1.14, but in another PR

antoineco: Fine with me. The current `alpine` tag is an alias for `1.14.1-alpine3.11` by the way (https://hub.docker.com/_/golang) so depending on when the last build occurred, we may already be using 1.14.

.dockerignore
sameersbn: good catch. it should list all the binaries. Will update the PR

awscodecommit/Dockerfile
sameersbn: `ENV` instructions no longer result in filesystem layers. Since Docker 1.10 only the instructions `RUN`, `COPY`, `ADD` create layers. In the interest of making the `Dockerfile` more readable i split those `ENV` and `LABEL` instructions

source: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers

awscodecommit/Dockerfile
sameersbn: @antoineco  same response as above :wink: 
@sebgoa thanks for the clarification

.dockerignore
sameersbn: there is `make coverage` target in the `Makefile` that generates `c.out` file which is input to `go tool` to generate `coverage.html`

awscodecommit/Dockerfile
sameersbn: i don't mind updating the golang version in this PR.

sameersbn: done

sameersbn:
> Comments apply to all sub-projects. Mostly curiosity.

Thanks for the review.

scripts/inc.Makefile
sameersbn: credits to @odacremolbap

awscodecommit/Dockerfile
antoineco: I missed that update, thanks for the pointers!

scripts/inc.Makefile
```diff
@@ -0,0 +1,53 @@
...
+	$(GO) mod download
+
+build: ## Build the binary using 'go build'
+	$(GO) build -ldflags "$(LDFLAGS)" -installsuffix cgo -o $(PACKAGE)
+
+install: ## Install the binary using the 'go install'
+	$(GO) install -ldflags "$(LDFLAGS)" -installsuffix cgo
+
+test: ## Run unit tests
+	$(GOTEST) -count=1 ./...
```
antoineco, line 30: From my experience `go test` caching is really clever and actually something you want to leverage as a developer. Any reason to disable it?

sameersbn: i was getting conflicting results while working on #47. Depending on how I was invoking the tests `make test` vs `make -C awscodecommit test` i was seeing different results :man_shrugging:

sameersbn: maybe I was doing something wrong. will remove it since it makes sense to have cached results

antoineco: Thanks! Let's investigate deeper in case it happens again.

PR #47, opened 2020-04-02
fixes broken tests

sameersbn:
The components require certain environment variable to exist in the environment. Certain tests fail because specific environment variables do no exist.

f.e.

```
--- FAIL: TestSendCloudevent (0.00s)
    main_test.go:148: 
        	Error Trace:	main_test.go:148
        	Error:      	Received unexpected error:
        	            	subject: if present, MUST be a non-empty string
        	Test:       	TestSendCloudevent
```

### DIFF
--- a/awscognito/main_test.go
+++ b/awscognito/main_test.go
@@ -136,6 +136,8 @@ func TestSendCognitoEvent(t *testing.T) {
 	httpmock.Activate()
 	defer httpmock.DeactivateAndReset()
 
+	identityPoolID = "fooPool"
+
 	httpmock.RegisterResponder("POST", "https://foo.com", httpmock.NewStringResponder(200, ``))
 	httpmock.RegisterResponder("POST", "https://bar.com", httpmock.NewStringResponder(500, ``))
 
--- a/awsdynamodb/main_test.go
+++ b/awsdynamodb/main_test.go
@@ -248,6 +248,8 @@ func TestSendCloudevent(t *testing.T) {
 	httpmock.Activate()
 	defer httpmock.DeactivateAndReset()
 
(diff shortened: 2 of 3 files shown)


a-edakin:
I think it is fine, you can merge

awscognito/main_test.go
```diff
@@ -136,6 +136,8 @@ func TestSendCognitoEvent(t *testing.T) {
 	httpmock.Activate()
 	defer httpmock.DeactivateAndReset()
 
+	identityPoolID = "fooPool"
```
antoineco, line 139: This value is read [from an environment variable](https://github.com/triggermesh/knative-lambda-sources/blob/5605f3a4da7a061da437a7aa7b7112500e49e36e/awscognito/main.go#L72) inside the main. I think the correct thing to do here is to initialize the env vars you need [inside your test](https://github.com/kyma-project/kyma/blob/master/components/event-sources/reconciler/httpsource/controller_test.go#L55-L56) to remain close to what the code you test actually does.

sameersbn: the issue is `main.go` picks up the environment values (global variables) in the `init` call, setting `os.Setenv` in the test would get called after `os.GetEnv`, so it's too late. 

I could possibly fetch the env variables in the calls instead of having global variables. Do you think it's the right thing to do or is there some alternate method?

https://github.com/triggermesh/knative-lambda-sources/blob/master/awscognito/main.go#L68-L76

antoineco: Right, I did not pay attention to the `init`.

A good practice would be to avoid `init` and package globals altogether, but that's not the topic of this PR, so I guess that approach is fine for the current code ✅

Issue #65, opened 2020-04-09
Add a structural validation schema to events source's CRDs


antoineco:
Mandatory to support `apiextensions.k8s.io/v1`.

See https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#specifying-a-structural-schema

There are a few good examples in Knative, like https://github.com/knative/eventing/blob/master/config/core/resources/channel.yaml

PR #76, opened 2020-04-14
Add AWSCognitoSource type and corresponding reconciler

### LINKED ISSUE #69: Create spec and reconciler for awscognito source

antoineco:
A shameless copy-paste of the AWS CodeCommit source, just with different attributes.

Closes #69


pkg/reconciler/awscognitosource/adapter.go
```diff
@@ -0,0 +1,228 @@
...
+	LoggingCfg string `ignored:"true"`
+	// Metrics (observability) configuration, serialized as JSON.
+	MetricsCfg string `ignored:"true"`
+}
+
+// reconcileAdapter reconciles the state of the source's adapter.
+func (r *Reconciler) reconcileAdapter(ctx context.Context,
+	src *v1alpha1.AWSCognitoSource) (*appsv1.Deployment, error) {
+
+	sinkRef := &src.Spec.Sink.Ref
```
tzununbekov, line 62: wouldn't it panic if there is no `Sink` in source object? Asking because in knative sources it can be the case and currently `Sink` is being verified in the webhook: https://github.com/knative/eventing-contrib/pull/1120#discussion_r406812534

pkg/reconciler/awscognitosource/controller.go
```diff
@@ -0,0 +1,80 @@
...
+	sourceInformer.Informer().AddEventHandler(controller.HandleAll(impl.Enqueue))
+
+	deploymentInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: controller.FilterGroupVersionKind((&v1alpha1.AWSCognitoSource{}).GetGroupVersionKind()),
+		Handler:    controller.HandleAll(impl.EnqueueControllerOf),
+	})
+
+	r.sinkResolver = resolver.NewURIResolver(ctx, impl.EnqueueKey)
+
+	cmw.Watch(logging.ConfigMapName(), r.updateAdapterLoggingConfig)
```
tzununbekov, line 76: I may be wrong, but logging config is handled on the shared controller level:
https://github.com/knative/eventing-contrib/issues/1067

same may be true for metrics config

pkg/reconciler/awscognitosource/adapter.go
antoineco: You're absolutely right. My intention is to implement the essential parts of the validation in #65 before the feature branch gets merged to `master`, so I didn't bother validating too many things inside the reconciliation code because I would most likely delete it afterwards.

pkg/reconciler/awscognitosource/controller.go
antoineco: That'd be great! But then where do you get the config from? In your code you set both to `""`, which doesn't seem right.

tzununbekov: I found where logging config comes from if adapter env doesn't have it:
https://github.com/knative/pkg/blob/master/logging/config.go#L160

but I'm wondering what was the need to set `required:"true"` parameter for `K_LOGGING_CONFIG` env variable and have default fallback in the same time

pkg/reconciler/awscognitosource/adapter.go
tzununbekov: Ok, makes sense

pkg/reconciler/awscognitosource/controller.go
antoineco: Nice catch. This will always use the default settings though, which means there is no possibility to enable debug logging for example.With the cmw, the changes are propagated down to adapters which is really neat.

Maybe we can find a way to have a _single_ watcher instead of one per reconciler, and propagate the logging config in the main context.

tzununbekov: I'm still trying to understand why I was asked to remove cmw from gitlabsource; I expected to see that watcher was moved to the pkg/adapter package, but it doesn't seem to be true. Maybe adapter's pod (kservice) lifecycle doesn't assume watching for logging updates... :confused: Anyways

> Maybe we can find a way to have a single watcher instead of one per reconciler, and propagate the logging config in the main context.

yeah, that's what I was thinking too

PR #97, opened 2020-04-16
set a subject to be compatible with argo sensors

sebgoa:
Just a quick one to discuss the use of CE subject.

If we could use a sane default it would make our sources compatible with argo-sensors.


antoineco:
I think it's sensible to have a subject. However we should not be using a constant string here. The [CloudEvents spec](https://github.com/cloudevents/spec/blob/master/spec.md#subject) has a good example of what a `subject` can be. In the case of a SQS queue, I believe `msg.MessageId` would be a better subject. I can add as part of #81 if you prefer.

antoineco:
ref #99

PR #105, opened 2020-04-17
build: test the pkgs only once

sameersbn:
added top-level Makefile targets to test pkgs


Makefile
```diff
@@ -10,3 +10,29 @@ $(SUBDIRS):
 
 # Code generation
 include scripts/Makefile.codegen
+
+include scripts/inc.Makefile
+
+GOPKGS = ./pkg/adapter/... ./pkg/reconciler/...
```
sameersbn, line 16: @antoineco should i change this to `./pkg/...` instead?

cmd/awscognitosource/Makefile
```diff
@@ -1,6 +1,5 @@
 PACKAGE = awscognitosource
 PACKAGE_DESC = Triggermesh AWS Cognito Event Source
 
-GOPKGS = ../../pkg/...
-
 include ../../scripts/inc.Makefile
+include ../../scripts/inc.Target
```
antoineco, line 5: Looks like your editor removes the final new line from some files, but not from others. In general I'd prefer to keep them (most editors do by default), because files without a final new line cause weird things in shell prompts when you `cat` them:
```console
acotten:/git/aws-event-sources $ cat cmd/awscognitosource/Makefile
PACKAGE = awscognitosource
PACKAGE_DESC = Triggermesh AWS Cognito Event Source

GOPKGS = ../../pkg/...

include ../../scripts/inc.Makefile
include ../../scripts/inc.Targetacotten:/git/aws-event-sources $
```

Makefile
antoineco, line 16: `./pkg/apis/...` will have few tests as well.

.dockerignore
```diff
@@ -15,8 +15,8 @@
 **/kodata/
 **/*_test.go
 **/c.out
-**/*-coverage.html
-**/*-unit-tests.xml
```
antoineco, line 19: From what I see the pattern still contains a dash?

Makefile
antoineco: Maybe just add `./pkg/apis/...`. Including `client/generated` would make your coverage really bad :D

cmd/awscognitosource/Makefile
sameersbn: Yea. I recently reset my editor. Need to configure it to add the new line

.dockerignore
sameersbn: fixed

PR #113, opened 2020-04-21
Update README

antoineco:
📄  [HTML render](https://github.com/antoineco/triggermesh-aws-event-sources/blob/update-readme/README.md)


sebgoa:
Good for now, I think we will need to have some basic usage instructions to run the released controller

antoineco:
That's true. Not sure if we want to release manifests together with the container images or leave that to the user, for example by replacing `image` attributes inside `ko` manifests. Open to discussions / suggestions.

PR #114, opened 2020-04-22
ci: adds `checkout` job and persists the results in the workspace

sameersbn:
The new checkout job checks out the git source and populate the go
modules. The results are then persisted in the workspace and used in
other jobs. This way we are not repeatedly doing `checkout + go mod
download` for every job

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,21 +5,26 @@ orbs:
   gcp-cli: circleci/gcp-cli@1.8.4
 
 jobs:
-  build:
+  checkout:
     executor:
       name: go/default
       tag: '1.14'
     steps:
       - checkout
-      - restore_cache:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -5,21 +5,26 @@ orbs:
...
     steps:
       - checkout
-      - restore_cache:
-          key: v1-repo-{{ .Environment.CIRCLE_BRANCH }}
-      - run:
-          name: Downloading Go modules
-          command: make mod-download
-      - save_cache:
-          key: v1-repo-{{ .Environment.CIRCLE_BRANCH }}
+      - go/mod-download-cached
```
sameersbn, line 14: since the `go.sum` and `go.mod` files now exist in the repo root, we can use the function from the golang orb which  cleans up some lines

Issue #27, opened 2020-02-27
Please support Kube2IAM auth, instead of cred file


AceHack:
I would like to use the credentials built into the POD via Kube2IAM instead of cred file

sebgoa:
For which source in particular ?

sebgoa:
Never tried but we could check that the Pod has a the kube2iam annotation and if yes then not try to load the secrets.

Set an env var to read the annotation from the metadata:

```
        - name: MY_POD_ANNOTATION
          valueFrom:
            fieldRef:
              fieldPath: metadata.annotation
```

and that in the pod template:

```
spec:
...
  template:
    metadata:
      annotations:
        iam.amazonaws.com/role: TEST_ROLE_NAME
```

antoineco:
Now that we have our own custom type per source, we could extend the API and provide a generic way to select an authentication method ((e.g. within the [`AWSSecurityCredentials` struct](https://github.com/triggermesh/aws-event-sources/blob/feature/reconciler-adapter/pkg/apis/sources/v1alpha1/common_types.go#L25-L28).

antoineco:
An alternative to Kube2iam worth considering, maybe: https://blog.pipetail.io/posts/2020-04-13-more-eks-tips/

AceHack:
Yes at my company we moved to AWS service account IAM roles recently and will be deprecating our use of Kube2IAM soon.

AceHack:
cred files would still be needed for non AWS deployment like azure and local.

antoineco:
OK thanks for the feedback @AceHack! Indeed, this only works in EKS. 
Doesn't Kube2iam work in any environment? If I understand the way it works correctly, pods shouldn't even be aware that the metadata endpoint they are talking to is provided by Kube2iam.

AceHack:
I personally didn't set up kube2iam.  But I thought it only worked in AWS.

bricelalu:
Hello,

I am currently using the [ServiceAccounts to give IAM role to my pods](https://www.eksworkshop.com/beginner/110_irsa/iam-role-for-sa-2/), I would like to do the same with the event-sources, how can I give a serviceAccount to a source ?

antoineco:
@bricelalu right now the controller manages the service account. As soon as you create an event source (of any type), a ServiceAccount named `<kind>-adapter` is created inside the same namespace as the source object (for example `awss3source-adapter`).

[...]

PR #116, opened 2020-04-22
Remove explicit handling of AWS credentials

### LINKED ISSUE #107: Remove explicit handling of AWS_* environment variables in adapters
The AWS SDK for Go already understands the following environment variables, so there is no need to handle them and store them in a variable within our adapters' code:
* AWS_ACCESS_KEY_ID
* AWS_SECRET_ACCESS_KEY
* AWS_REGION

see https://docs.aws.amazon.com/sdk-for-go/api/aws/session/#hdr-Environment_Variables

antoineco:
`NewSession` loads credentials automatically from multiple sources in [order of preference](https://docs.aws.amazon.com/sdk-for-go/api/aws/session/). Letting it handle this order gives us more flexibility to use credentials files (local) or even EC2 metadata (kube2iam).

Ref #27
Closes #107

### DIFF
--- a/pkg/adapter/awscodecommitsource/adapter.go
+++ b/pkg/adapter/awscodecommitsource/adapter.go
@@ -24,7 +24,6 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/codecommit"
 	"github.com/aws/aws-sdk-go/service/codecommit/codecommitiface"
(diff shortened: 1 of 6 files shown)


pkg/adapter/awscodecommitsource/adapter.go
```diff
@@ -87,27 +82,18 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor,
...
 	// create CodeCommit client
-	sess, err := session.NewSession(&aws.Config{
-		Region:      &env.AWSRegion,
-		Credentials: credentials.NewStaticCredentials(env.AccountAccessKeyId, env.AccountSecretAccessKey, ""),
-		MaxRetries:  aws.Int(5),
-	})
-	if err != nil {
-		logger.Fatalw("Failed to create CodeCommit client", "error", err)
-	}
+	sess := session.Must(session.NewSession(aws.NewConfig().WithMaxRetries(5)))
```
sameersbn, line 85: Do we need to specify the region with `aws.NewConfig()`?

```suggestion
	sess := session.Must(session.NewSession(aws.NewConfig().WithRegion(env.AWSRegion).WithMaxRetries(5)))
```

antoineco: The region is also set from the environment (`AWS_REGION`). If this env var is missing, the client falls back to reading your local config, which is quite handy when debugging sources locally. (e.g. I can have it default to eu-central and you to us-west without any explicit parameter for example)

sameersbn: thanks for the clarification

PR #123, opened 2020-04-29
build improvements

sameersbn:
- ci: use env variables from org contexts
- add release binaries to git ignore rules
- test: increase verbosity of `gotestsum` failures
- docker: exclude release binaries from docker build context
- lint: enforce linter errors


sameersbn:
@antoineco small set of improvements separated by commit

PR #124, opened 2020-04-29
Use empty logging config in docs

antoineco:
The previous value didn't have the correct format, and a Zap configuration needs to contain [much more than just a loglevel](https://github.com/knative/pkg/blob/release-0.14/logging/config.go#L132-L151) to be parsed successfully. (e.g. `encoderConfig` must contain all the required attributes)

Instead of suggesting users to set gigantic and unreadable env vars, I suggest we use an empty string in docs and samples, which would fall back to the defaults linked above.

### DIFF
--- a/cmd/awscodecommitsource/README.md
+++ b/cmd/awscodecommitsource/README.md
@@ -86,7 +86,7 @@ export AWS_ACCESS_KEY_ID=<my_key_id>
 export AWS_SECRET_ACCESS_KEY=<my_secret_key>
 export NAME=my-awscodecommitsource
 export NAMESPACE=default
-export K_LOGGING_CONFIG='{"level":"info"}'
+export K_LOGGING_CONFIG=''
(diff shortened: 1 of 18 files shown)


sebgoa:
easy enough

/lgtm

antoineco:
For posterity, here's what the users would have to export in order to configure Zap to their needs:
```sh
K_LOGGING_CONFIG='{"zap-logger-config":"{ \"level\": \"debug\", \"development\": false, \"outputPaths\": [\"stdout\"], \"errorOutputPaths\": [\"stderr\"], \"encoding\": \"json\", \"encoderConfig\": {   \"timeKey\": \"ts\",   \"levelKey\": \"level\",   \"nameKey\": \"logger\",   \"callerKey\": \"caller\",   \"messageKey\": \"msg\",   \"stacktraceKey\": \"stacktrace\",   \"lineEnding\":\"\",   \"levelEncoder\": \"\",   \"timeEncoder\": \"iso8601\",   \"durationEncoder\":\"\",   \"callerEncoder\": \"\" }}"}'
```

sebgoa:
u shoudl tweet that :)

PR #132, opened 2020-05-06
Update dependencies

antoineco:
No functional change. Knative recently switched to Go modules, so bumping versions also simplifies our own dependencies.

I ran `make codegen` in the process to keep things up-to-date.


go.sum
```diff
@@ -1,389 +1,903 @@
...
 cloud.google.com/go/datastore v1.0.0/go.mod h1:LXYbyblFSglQ5pkeyhO+Qmw7ukd3C+pD7TKLgZqpHYE=
+cloud.google.com/go/datastore v1.1.0/go.mod h1:umbIZjpQpHh4hmRpGhH4tLFup+FVzqBi1b3c64qFpCk=
+cloud.google.com/go/logging v1.0.0/go.mod h1:V1cc3ogwobYzQq5f2R7DS/GvRIrI4FKj01Gs5glwAls=
+cloud.google.com/go/pubsub v1.0.1/go.mod h1:R0Gpsv3s54REJCy4fxDixWD93lHJMoZTyQ2kNxGRt3I=
+cloud.google.com/go/pubsub v1.1.0/go.mod h1:EwwdRX2sKPjnvnqCa270oGRyludottCI76h+R3AArQw=
+cloud.google.com/go/pubsub v1.2.0/go.mod h1:jhfEVHT8odbXTkndysNHCcx0awwzvfOlguIAii9o8iA=
+cloud.google.com/go/storage v1.0.0/go.mod h1:IhtSnM/ZTZV8YYJWCY8RULGVqBDmpoyjwiyrjsg+URw=
+cloud.google.com/go/storage v1.5.0/go.mod h1:tpKbwo567HUNpVclU5sGELwQWBDZ8gh0ZeosJ0Rtdos=
+cloud.google.com/go/storage v1.6.0/go.mod h1:N7U0C8pVQ/+NIKOBQyamJIeKQKkZ+mxpohlUTyfDhBk=
+contrib.go.opencensus.io/exporter/aws v0.0.0-20181029163544-2befc13012d0/go.mod h1:uu1P0UCM/6RbsMrgPa98ll8ZcHM858i/AD06a9aLRCA=
```
antoineco, line 36: I have no idea why Go mod adds so many entries to `go.sum`, event after running `go mod tidy`. I checked my cache, and there is actually only one version of each package in here. What a mystery..

sebgoa:
just wait couple years, there will be a replacement to go mod

antoineco:
_Go_ will end up on https://killedbygoogle.com/ just like the rest you mean!

antoineco:
`golangci-lint` always fails the first run, but succeeds on all subsequent runs. Reported at golangci/golangci-lint#1071

antoineco:
@sameersbn the linter is unhappy with those changes but I don't get why, I'm unable to reproduce locally 🤔  I've tried everything on the CircleCI side: disable cache entirely, checkout before running lint, run `mod-download-cached` before lint, ...

sameersbn:
taking a look

antoineco:
Maybe one hint I'm looking into: ` go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.26.0` alters the `go.sum` file. Maybe we should fetch the binary from GitHub releases instead. I still need to confirm that.

sameersbn:
I was testing the same thing. appears to be that is the issue!

antoineco:
Adding this for posterity:

> **How to use golangci-lint in CI (Continuous Integration)?**
>
> Run golangci-lint in CI and check the exit code. If it's non-zero - fail the build.
>
> We don't recommend vendoring golangci-lint in your repo: you will get troubles updating golangci-lint. Please, use recommended way to install with the shell script: it's very fast.

sameersbn:
funny thing is that the ci cannot find golangci-lint after the script installed it.

antoineco:
That's because by default it gets installed to `./bin`, which is not in the PATH. I just changed that.

Maybe we could also stick to the default, add `bin/` to all the `.*ignore` files, and avoid switching directories in the Makefile, since event sources are not independent modules as they used to be.

sameersbn:
could have just installed it in the GOPATH/bin

```
curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.26.0
```

PR #135, opened 2020-05-13
Update logging/metrics config using ConfigWatcher

antoineco:
Eventing has now a helper for configuring `ConfigMapWatcher` callbacks for logging/metrics/tracing configs, so we can get rid of all the redundant `updateAdapter...Config` methods.

The usage of a static `ConfigMapWatcher` is also not necessary anymore since the `source.ConfigAccessor` interface can be mocked.


sebgoa:
so if we merge this, this feature won't work with an old knative release ?

antoineco:
It will, this only makes use of internal eventing libraries, the clients work across versions within a range of supported API versions.

sebgoa:
ok

PR #138, opened 2020-05-15
Include controller in CloudBuilds

### LINKED ISSUE #112: Make aws-event-sources-controller part of releases
We should consider releasing a manifest file for the controller that would point to the release images for all sources, so that users can simply apply that manifest in their environment.

As part of this, we should also add some installation instructions to the README for running the controller.

See discussion in triggermesh/aws-event-sources#113

antoineco:
Makes the controller part of sources' releases.

Closes #112

### DIFF
--- a/cmd/aws-event-sources-controller/Dockerfile
+++ b/cmd/aws-event-sources-controller/Dockerfile
@@ -1,0 +1,20 @@
+FROM golang:1.14-alpine AS builder
+
+ENV CGO_ENABLED 0
+ENV GOOS linux
+ENV GOARCH amd64
+
+WORKDIR /go/src/aws-event-sources-controller
+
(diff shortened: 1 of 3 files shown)


cmd/aws-event-sources-controller/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM golang:1.14-alpine AS builder
+
+ENV CGO_ENABLED 0
+ENV GOOS linux
+ENV GOARCH amd64
+
+WORKDIR /go/src/aws-event-sources-controller
+
+RUN apk add --no-cache openssh-client ca-certificates
```
sameersbn, line 9: still trying to figure out why we want `openssh-client`. are the go modules that are downloaded with a git+ssh clone operation?

PR #142, opened 2020-05-24
Consolidate all reconcilers into generic Deployment reconcilers

antoineco:
Introduces a generic `AWSEventSource` interface which all AWS event sources implement, and a `GenericDeploymentReconciler` type which can reconcile that interface.

Since all AWS event sources use some identical reconciliation logic, this allows us to strip typed `ReconcileKind` functions to the bare minimum and avoid duplication.

Reconciliation tests will also be consolidated in the same manner in a follow-up PR.


pkg/reconciler/awscodecommitsource/reconciler.go
```diff
@@ -19,70 +19,29 @@ package awscodecommitsource
...
-	return r.reconcileAdapter(ctx, arn)
+func (r *Reconciler) ReconcileKind(ctx context.Context, src *v1alpha1.AWSCodeCommitSource) reconciler.Event {
+	// inject source into context for usage in reconciliation logic
+	ctx = v1alpha1.WithSource(ctx, src)
+
+	return r.base.ReconcileSource(ctx,
+		src.Spec.EventTypes,
+		adapterDeploymentBuilder(src, r.adapterCfg),
+	)
 }
```
antoineco, line 47: It now takes a single call to the generic `ReconcileSource` method in order to reconcile any AWS event source in the same manner.

pkg/reconciler/awscognitosource/adapter.go
```diff
@@ -50,128 +43,45 @@ type adapterConfig struct {
 	configs source.ConfigAccessor
 }
 
-// reconcileAdapter reconciles the state of the source's adapter.
-func (r *Reconciler) reconcileAdapter(ctx context.Context, arn arn.ARN) error {
```
antoineco, line 54: All the type-specific `reconcileAdapter` methods are now gone, replaced by a single generic reconciler.

PR #143, opened 2020-05-25
Build improvements

### LINKED ISSUE #141: Align CI resources on other repos
Also use the occasion to remove the usage of  `-installsuffix cgo` in builds.

antoineco:
PR contains the following changes:

* Remove usage of `-installsuffix cgo` build flag 
    Closes #141

* Omit debugging tables in builds for smaller binaries
    `-s`   Omit the [symbol table](https://en.wikipedia.org/wiki/Symbol_table) and debug information..
    `-w`   Omit the [DWARF](https://en.wikipedia.org/wiki/DWARF) symbol table.

* Generate individual make goals for container images by appending `.image` to each command
    `make aws-event-sources-controller.image`
    `make awscodecommitsource.image`
    ...


antoineco:
@sameersbn CircleCI is killing processes again, this time during tests. I find it a bit surprising considering the size of the machines. 2 CPU with 4 GB of memory is plenty!

sameersbn:
I think `make` is spawning more processes than the vcpu's the build machine is allotted. Can you try `make -j2 test` in the CI config?

sameersbn:
I remember we saw this earlier. is it in the same repo?

cmd/aws-event-sources-controller/Dockerfile
```diff
@@ -1,20 +1,29 @@
...
-RUN apk add --no-cache openssh-client ca-certificates
+COPY go.mod go.sum ./
+RUN go mod download
 
-COPY ./ .
-RUN go install -installsuffix cgo ./cmd/aws-event-sources-controller
+COPY . .
+RUN BIN_OUTPUT_DIR=/bin make aws-event-sources-controller && \
+    mkdir /kodata && \
+    mv .git/* /kodata/ && \
```
sameersbn, line 15: Honestly I am not sure what's going on with the `kodata` folder. Should be added for the other repo's as well?

antoineco:
Right, I remember now, you did set `-p=1`. It just pushed that change again, it got overridden in this PR (my bad).

Makefile
```diff
@@ -1,36 +1,119 @@
...
 
-clean:
-	$(RM) -rf $(PACKAGE)-unit-tests.xml
-	$(RM) -rf c.out $(PACKAGE)-coverage.html
+images: $(IMAGES) ## Builds container images
+
+$(IMAGES): %.image:
+	$(DOCKER) build -t $(IMAGE_REPO)/$* -f ./cmd/$*/Dockerfile . ;
+
+cloudbuild-test: ## Test container image build with Google Cloud Build
```
sameersbn, line 94: we should use this in the CI

cmd/aws-event-sources-controller/Dockerfile
antoineco: The Knative logger falls back to not logging the current SHA/tag if that's missing, so kodata is optional.
I personally like being able to filter logs by code revision in Kibana (that's what that feature is for), so it could make sense to have it by default on the long run.

Makefile
```diff
@@ -1,36 +1,118 @@
...
-	$(RM) -rf c.out $(PACKAGE)-coverage.html
+IMAGES = $(foreach cmd,$(COMMANDS),$(cmd).image)
+images: $(IMAGES) ## Builds container images
+$(IMAGES): %.image:
+	$(DOCKER) build -t $(IMAGE_REPO)/$* -f ./cmd/$*/Dockerfile . ;
+
+CLOUDBUILD_TEST = $(foreach cmd,$(COMMANDS),$(cmd).cloudbuild-test)
+cloudbuild-test: $(CLOUDBUILD_TEST) ## Test container image build with Google Cloud Build
+$(CLOUDBUILD_TEST): %.cloudbuild-test:
+	gcloud builds submit $(BASE_DIR) --config cmd/$*/cloudbuild.yaml --substitutions COMMIT_SHA=${IMAGE_SHA},_KANIKO_IMAGE_TAG=_
```
antoineco, line 96: @sameersbn in the previous version of this goal you were setting the `_KANIKO_NO_PUSH` substitution to `true` instead of setting `_KANIKO_IMAGE_TAG` to `_`. In fact, the underscore does not prevent the push and creates an actual `gcr.io/foo/bar:_` image.

What should we be using in test goals?

cmd/aws-event-sources-controller/cloudbuild.yaml
```diff
@@ -0,0 +1,28 @@
...
+substitutions:
+  _CMD: aws-event-sources-controller
+  _KANIKO_IMAGE_TAG: latest
+  _KANIKO_NO_PUSH: 'false'
+  _KANIKO_USE_BUILD_CACHE: 'true'
+  _KANIKO_EXTRA_ARGS: ''
+
+options:
+  substitution_option: ALLOW_LOOSE
+  machineType: N1_HIGHCPU_8
```
antoineco, line 25: @sameersbn this doesn't seem to be set in every project, do you remember if there was a reason for using 8-CPU machines here?

Makefile
antoineco: Done

cmd/aws-event-sources-controller/cloudbuild.yaml
sebgoa: I used that early on to be able to speed up the build. That was when the builds were serial and would time out after 10 minutes.

antoineco: OK, I'm removing it then. Builds shouldn't take _that_ long, even on a single CPU.

Makefile
sameersbn: @antoineco it's a hack. Earlier specifying `--no-push` on kaniko would not create a tag on gcr yet it would update the cache layers which was perfect. Something changed recently and the cloudbuild stated erroring out with authentication errors when  `--no-push` was being specified. Pushing the image with the `_` tag is my hack to update the cache to speed up subsequent builds.

cmd/aws-event-sources-controller/cloudbuild.yaml
sameersbn: @antoineco Since we've split it in separate cloudbuild.yaml files we should not need this. However I would set a higher build timeout to avoid failures due to slow network operations of degraded CPU performance from the build machines.

.circleci/config.yml
```diff
@@ -32,8 +32,7 @@ jobs:
       - gcp-cli/initialize
       - run:
           name: Building docker image
-          command: |
-            gcloud builds submit --config cloudbuild.yaml --substitutions _KANIKO_NO_PUSH=true
+          command: make -j8 cloudbuild-test
```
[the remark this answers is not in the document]
sameersbn: It was necessary to add it, otherwise only one build was being processed at a time.

Makefile
antoineco: OK thanks! I just added this as a comment in the Makefile so we don't forget why we do things that way.

PR #147, opened 2020-05-27
AWSCognitoUserPoolSource

tzununbekov:


cmd/awscognitouserpoolsource/kodata/HEAD
```diff
@@ -0,0 +1 @@
+ref: refs/heads/master
```
antoineco, line 1: Those should be symlinks. You can copy these from another cmd directory using the `-P` option:
```console
$ cp -rP cmd/aws-event-sources-controller/kodata cmd/awscognitouserpoolsource/
```

tzununbekov: missed this, thanks

config/samples/awscognitouserpool-containersource.yaml
```diff
@@ -0,0 +1,67 @@
...
+spec:
+  template:
+    spec:
+      containers:
+      - name: source
+        image: gcr.io/triggermesh/awscognitouserpoolsource:latest
+
+        env:
+
+        # Cognito identity pool
```
antoineco, line 30: "user pool"?

tzununbekov: copy-paste detected :rotating_light: 
thanks

pkg/reconciler/awscognitouserpoolsource/controller_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+package awscognitouserpoolsource
+
+import (
+	"testing"
+
+	. "github.com/triggermesh/aws-event-sources/pkg/reconciler/testing"
+
+	// Link fake informers accessed by our controller
+	_ "github.com/triggermesh/aws-event-sources/pkg/client/generated/injection/informers/sources/v1alpha1/awscognitosource/fake"
```
antoineco, line 25: This magic import needs to be updated.

pkg/adapter/awscognitouserpoolsource/adapter_test.go
```diff
@@ -0,0 +1,99 @@
...
+			},
+		},
+	}
+
+	go func() {
+		err := a.Start(nil)
+		assert.NoError(t, err)
+	}()
+	// Wait while first UserType array processed
+	time.Sleep(3 * time.Second)
```
antoineco, line 91: I'm sure we can do better than that by receiving a notification on a channel once the processing completes. I can check tomorrow.

antoineco: Maybe we should just test start/stop separately, and just test the handler here, unless you just want to provide a single high level test. That's what I did in the EventBridge target.

tzununbekov: agree, test looks ugly. I'll work on improvement tomorrow

tzununbekov: @antoineco it looks like you don't use `sharedmain` package for adapter in EventBridge, right? 
I've tried to move code from Start() to make it a bit more testable. It ain't much, but it's honest work :man_farmer:

antoineco: That's right, I don't use the shared main but it should be possible to test `Start()` in the same way by simply initializing you adapter struct. I'll take another look.

pkg/adapter/awscognitouserpoolsource/adapter_test.go
```diff
@@ -0,0 +1,150 @@
...
+func (m mockedCognitoUserPoolClient) ListUsers(in *cognitoidentityprovider.ListUsersInput) (*cognitoidentityprovider.ListUsersOutput, error) {
+	return &m.listUsersOutput, nil
+}
+
+func TestListUsers(t *testing.T) {
+	user := cognitoidentityprovider.UserType{
+		UserLastModifiedDate: func() *time.Time {
+			now := time.Now()
+			return &now
+		}(),
```
antoineco, line 46: Using the `aws` [helper package](https://docs.aws.amazon.com/sdk-for-go/api/aws/#Time):
```suggestion
		UserLastModifiedDate: aws.Time(time.Now())
```

pkg/adapter/awscognitouserpoolsource/adapter_test.go
```diff
@@ -0,0 +1,150 @@
...
+	// we must get 0 new users
+	users, timestamp := filterByTimestamp([]*cognitoidentityprovider.UserType{
+		&user1,
+	}, now)
+	assert.Len(t, users, 0)
+
+	// new user
+	user2 := cognitoidentityprovider.UserType{
+		UserLastModifiedDate: func() *time.Time {
+			now := time.Now()
```
antoineco, line 84: nit: This works because the code is sequential and the time is down to the nanosecond, but as a general practice in tests you could also fake timestamps with [`time.Unix(0, 0)`](https://golang.org/pkg/time/#Unix) (and increment manually) to ensure they work in any situation.

pkg/adapter/awscognitouserpoolsource/adapter_test.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	err := a.sendCognitoEvent(&user)
+	assert.NoError(t, err)
+
+	events := ceClient.Sent()
+	assert.Len(t, events, 1)
+
+	var gotUser cognitoidentityprovider.UserType
+	err = events[0].DataAs(&gotUser)
```
antoineco, line 128: Nice, I didn't know you could do that!

pkg/adapter/awscognitouserpoolsource/adapter_test.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	stopCh := make(chan struct{})
+	go func() {
+		close(stopCh)
+	}()
+
+	// test graceful shutdown
+	err := a.Start(stopCh)
+	assert.NoError(t, err)
```
antoineco, line 149: Your `Start()` function may never react to the signal, and then your test will run until it times out (10 min). Here's a suggestion to ensure we time out quickly (after 2s):

```go
	// create a context that will be done in 2s
	testCtx, testCancel := context.WithTimeout(context.Background(), 2*time.Second)
	defer testCancel()  // not really necessary here, but let's follow good practices

	// channel to receive stop signals
	stopCh := make(chan struct{})

	// channel to receive the value returned by Start()
	errCh := make(chan error)

	// run adapter handler in the background
	go func() {
		errCh <- a.Start(stopCh)
	}()

	// simulate sending a stop signal to the adapter
	close(stopCh)

	// if the timeout is done before the handler returned, the test is always failed
	select {
	case <-testCtx.Done():
		t.Errorf("Test timed out after %v", testTimeout)
	case err := <-errCh:
		assert.NoErrorf(t, err, "Receiver returned an error")
	}
```

pkg/adapter/awscognitouserpoolsource/adapter_test.go
```diff
@@ -0,0 +1,162 @@
...
+	a := &adapter{
+		userPoolID:        "userpool/fooPool",
+		logger:            loggingtesting.TestLogger(t),
+		ceClient:          adaptertest.NewTestClient(),
+		cgnIdentityClient: mockedCognitoUserPoolClient{},
+	}
+
+	// create a context that will be done in 2s
+	testCtx, testCancel := context.WithTimeout(context.Background(), testTimeout)
+	defer testCancel() // not really necessary here, but let's follow good practices
```
antoineco, line 139: That comment was for the review 🙂   It can stay but it's a little bit unnecessary.

pkg/adapter/awscognitouserpoolsource/adapter_test.go
```diff
@@ -0,0 +1,162 @@
...
+	testTimeout := 2 * time.Second
+
+	a := &adapter{
+		userPoolID:        "userpool/fooPool",
+		logger:            loggingtesting.TestLogger(t),
+		ceClient:          adaptertest.NewTestClient(),
+		cgnIdentityClient: mockedCognitoUserPoolClient{},
+	}
+
+	// create a context that will be done in 2s
```
antoineco, line 137:
```suggestion
	// create a context that will be done after testTimeout
```

pkg/adapter/awscognitouserpoolsource/adapter_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	var gotUser cognitoidentityprovider.UserType
+	err = events[0].DataAs(&gotUser)
+	assert.NoError(t, err)
+	// make sure that we get what we sent
+	assert.Equal(t, user, gotUser)
+}
+
+func TestStart(t *testing.T) {
+	testTimeout := 2 * time.Second
```
antoineco, line 128:
```suggestion
	const testTimeout = 2 * time.Second
```

tzununbekov:
@antoineco thanks for these useful little tips and patience :bow:

PR #151, opened 2020-06-01
Dynamodb source cleaned up

### LINKED ISSUE #140: DynamoDB: starts even if the table or stream does not exist
The DynamoDB source should fail early when the table or table stream do not exist. Right now it starts without providing any feedback.

tzununbekov:
- fail if table or streams doesn't exist, closes #140
- unused waitgroups removed
- events handling moved out of goroutines
- encode whole dynamodb record into cloudevent
- ~~record sequence id as CE subject~~

TODO
- add poll backoff


pkg/adapter/awsdynamodbsource/adapter.go
```diff
@@ -209,42 +202,24 @@ func (a *adapter) processLatestRecords(shardIterator *string) error {
...
-		EventName:    record.EventName,
-		EventSource:  record.EventSource,
-		EventVersion: record.EventVersion,
-		UserIdentity: record.UserIdentity,
-	}
-
 	event := cloudevents.NewEvent(cloudevents.VersionV1)
 	event.SetType(v1alpha1.AWSEventType(a.arn.Service, strings.ToLower(*record.EventName)))
-	event.SetSubject(asEventSubject(record))
+	event.SetSubject(*record.Dynamodb.SequenceNumber)
```
antoineco, line 219: I implemented that weird looking function because setting the sequence number as the subject wasn't very much aligned with the description of a "subject" from the CloudEvents API.

I'm not sure whether my approach is the best we could come up with (using the name of the modified key(s) as subject), but I believe we should stay away from cryptic numbers in "subject" attributes.

There is another issue opened to describe this kind of things and what people should expect from our events: what convention do we use for "source" or "subject" for example, or what extensions we set, if any.

pkg/adapter/awsdynamodbsource/events.go
```diff
@@ -1,32 +0,0 @@
...
-*/
-
-package awsdynamodbsource
-
-import (
-	"github.com/aws/aws-sdk-go/service/dynamodbstreams"
-)
-
-// DynamoDBEvent represents an AWS Dynamo DB payload.
-type DynamoDBEvent struct {
```
antoineco, line 24: Nice! I was wondering why we needed that :)

tzununbekov: yep, I think we can do that for the rest of adapters as well

pkg/adapter/awsdynamodbsource/adapter.go
tzununbekov: > we should stay away from cryptic numbers

agree. I'm totally okay with restoring subjects
I didn't find any description of "subject" field and thought that it could be an opportunity to simplify the code, especially considering that event body now contains full DynamoDB record including keys and we can get rid of additional calculations on our side and leave it to the receiver

PR #152, opened 2020-06-02
Exponential backoff for receiver poll loops

### LINKED ISSUE #139: Agressive polling causes rate limit hits
All AWS sources poll AWS APIs in a `for` loop, which causes a lot of issues due to rate limiting.

At the very least we should throttle our polling with a [`time.Ticker`](https://golang.org/pkg/time/#Ticker).

tzununbekov:
- got rid of events custom type in favor of complete AWS event body
- adapter's `Start()` method uses `backoff.Run()` function to invoke receiver logic with exponential backoff duration, closes #139


pkg/adapter/common/backoff.go
```diff
@@ -0,0 +1,107 @@
...
+		min:    minBackoff,
+		max:    maxBackoff,
+	}
+
+	switch len(args) {
+	case 1:
+		backoff.min = args[0]
+	case 2:
+		backoff.min = args[0]
+		backoff.max = args[1]
```
antoineco, line 55: We should assert that `backoff.min <= backoff.max` and just ignore if it's not the case, maybe along with a small test.

pkg/adapter/common/backoff.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	return &backoff
```
antoineco, line 58: nit: You can also initialize `backoff` as a pointer to struct directly at line 43.

pkg/adapter/common/backoff.go
```diff
@@ -0,0 +1,107 @@
...
+	default:
+		b.step++
+		return dur
+	}
+}
+
+// Run is a blocking function that executes fn(ctx) until stopCh receives the value or fn returns an error.
+// fn(ctx) must return (bool, error) values where bool is true if backoff counter must be reset
+// and error is the result of fn() execution.
+func (b *Backoff) Run(stopCh <-chan struct{}, fn func(context.Context) (bool, error)) error {
```
antoineco, line 80: It would make things clearer if we had a type definition for `func() func(context.Context) (bool, error)`, e.g.
```go
type RunFunc func(context.Context) (bool, error)

func (b *Backoff) Run(stopCh <-chan struct{}, fn RunFunc) error
```

and then in use it as follows in callers:
```go
err := backoff.Run(stopCh, backoff.RunFunc {
	return func(ctx context.Context) (bool, error) { // compiler will implicitly assert as RunFunc
 		// ...
	}
}
```

pkg/adapter/awscodecommitsource/adapter.go
```diff
@@ -140,32 +140,44 @@ func (a *adapter) Start(stopCh <-chan struct{}) error {
...
-			}
-		}
+	backoff := common.NewBackoff()
 
-		if strings.Contains(a.gitEvents, prEventType) {
-			pullRequests, err := a.preparePullRequests()
-			if err != nil {
-				a.logger.Errorw("Failed to process pull requests", "error", err)
-			}
+	err = backoff.Run(stopCh, func() func(context.Context) (bool, error) {
```
antoineco, line 145: I'm having a hard time wrapping my head around that `func() func(context.Context) (bool, error) { /* */ }()`. Why don't we pass a `func(context.Context) (bool, error)` directly?

tzununbekov: We do now, thanks!

pkg/adapter/awssqssource/adapter.go
```diff
@@ -84,56 +83,49 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClie
...
-
-		// Delete message from queue if we pushed successfully
-		err = a.deleteMessage(queueURL, msgs[0].ReceiptHandle)
-		if err != nil {
-			a.logger.Errorw("Failed to delete message from SQS queue", "error", err)
-			continue
+	backoff := common.NewBackoff()
+
+	err = backoff.Run(stopCh, func() func(context.Context) (bool, error) {
+		return func(ctx context.Context) (bool, error) {
```
antoineco, line 100: This one still uses the "complex" signature, is that on purpose?

tzununbekov: no, it isn't :see_no_evil:

PR #153, opened 2020-06-03
Consolidate reconciler tests for all sources

antoineco:
D.R.Y. effort to avoid repeating the same test code in every source.

Part two of #142


pkg/reconciler/testing/reconcile.go
```diff
@@ -0,0 +1,392 @@
...
+func TestReconcile(t *testing.T, ctor Ctor, src v1alpha1.AWSEventSource,
+	eventTypes []string, adapterFn AdapterDeploymentBuilderFuncForSource) {
+
+	arn, err := arn.Parse(src.GetARN())
+	if err != nil {
+		t.Fatalf("provided source object has invalid ARN: %s", err)
+	}
+
+	newEventSource := eventSourceCtor(src, arn, eventTypes)
+	newAdapterDeployment := adapterDeploymentCtor(adapterFn(newEventSource()), arn)
```
antoineco, line 73: I'm not proud of this mess.

The whole complexity of this series of function calls comes from the fact that we need to "prepare" the given source with a call to `eventSourceCtor()` before we can use it to generate an adapter Deployment. The reason for that is that `eventSourceCtor()` populates the source's metadata (namespace, name, uid) with constant attributes, and that metadata plays in role in the generation of the adapter Deployment (name, owner reference, ...).

Maybe it would be better to populate these _before_ passing the source to the test function. That would at least eliminate the call to `newEventSource()`.

pkg/reconciler/testing/reconcile.go
```diff
@@ -0,0 +1,392 @@
...
+
+	testCases.Test(t, MakeFactory(ctor))
+}
+
+/* Event sources */
+
+// eventSourceCtor creates a copy of the given source and returns a function
+// that can apply options to that sources.
+func eventSourceCtor(src v1alpha1.AWSEventSource, arn arn.ARN,
+	eventTypes []string) func(...sourceOption) v1alpha1.AWSEventSource {
```
antoineco, line 224: I have an idea to simplify this a bit.

The source contains everything that's needed to get the ARN and event types, so I might as well pass just the source and add a new `GetEventTypes()` method to the AWSEventSource interface method to fetch the event types. I just wanted to avoid touching the interface in this PR.

pkg/reconciler/testing/reconcile.go
tzununbekov: will it be helpful if we move `adapterDeploymentBuilder` from each source to common package and expose methods to set source-specific specs?

antoineco: Maybe, although it's not very "common" anymore in this case. I'll see how that could look like.

pkg/reconciler/testing/reconcile.go
```diff
@@ -0,0 +1,392 @@
...
+
+// Test the Reconcile method of the controller.Reconciler implemented by controllers.
+//
+// The environment for each test case is set up as follows:
+//  1. MakeFactory initializes fake clients with the objects declared in the test case
+//  2. MakeFactory injects those clients into a context along with fake event recorders, etc.
+//  3. A Reconciler is constructed via a Ctor function using the values injected above
+//  4. The Reconciler returned by MakeFactory is used to run the test case
+func TestReconcile(t *testing.T, ctor Ctor, src v1alpha1.AWSEventSource,
+	eventTypes []string, adapterFn AdapterDeploymentBuilderFuncForSource) {
```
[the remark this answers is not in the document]
antoineco: The eventTypes param was removed, it is now read directly from the source.

pkg/reconciler/testing/reconcile.go
antoineco: I simplified those functions.

1. All sources must now be sent pre-populated to `TestReconcile`, so the call to `newEventSource` was removed.
2. `adapterDeploymentCtor` now receives `adapterFn, src`, which allows us to call `adapterFn(src)` _and_ read the ARN from `src`.

pkg/reconciler/testing/reconcile.go
antoineco: Done

Issue #155, opened 2020-06-10
Add AWS Cognito UserPool to readme


tzununbekov:

tzununbekov:
covered during Cognito sources rename:
https://github.com/triggermesh/aws-event-sources#list-of-event-sources

Issue #148, opened 2020-05-29
Fix AWS Cognito source (Identity pools)


tzununbekov:
test current AWS Cognito Identity Pool source, fix if needed or replace with User Pool source

tzununbekov:
Couldn't make current AWS Cognito source working, ended up by replacing with new AWS CognitoUserPoolSource

sebgoa:
I don't understand. We have a Cognito source already. Are you saying it does not work ? if yes we should fix it.

And/or are you saying that the CognitoUserPoolSource is a new source that does the same thing as the previous Cognito source ?

tzununbekov:
I just needed to have an issue to reflect my work on AWS Cognito. Besides, after some investigation I had an impression that User Pool is more common case for AWS Cognito event source while Identity Pool requires either [AWS Mobile or JavaScript SDK](https://docs.aws.amazon.com/cognito/latest/developerguide/getting-started-with-identity-pools.html#install-the-mobile-or-javascript-sdk) to work with it. That is, until we figure out the flows to use AWS Cognito Identity Pool, I'd call User Poll our default AWS Cognito event source. Anyway, I'm sorry that I confused everyone with this issue

antoineco:
Shall we take action and remove or rename the old source in that case?

tzununbekov:
If we worry that Triggermesh users get confused by Cognito/CognitoUserPool sources, then I'd rename them, although I don't have a strong opinion about that

PR #156, opened 2020-06-11
Rename AWS Cognito event sources

tzununbekov:
`AWSCognitoSource` event source renamed to `AWSCognitoIdentitySource`

related to #148, #155


antoineco:
I'd keep the explicit `AWSCognitoUserPoolSource` name to avoid confusion.

pkg/apis/sources/v1alpha1/awscognito_defaults.go
```diff
@@ -19,4 +19,4 @@ package v1alpha1
 import "context"
```
antoineco, line 19: Files should also be renamed.

Issue #157, opened 2020-06-12
Reconcile SNS subcriptions outside of the receive adapter


antoineco:
There is a lot more we could do in terms of reconciliation of SNS subscriptions if we were reconciling them in the controller instead of the adapter:

 * expose the status of the subscription (`PendingConfirmation`: True/False)
 * fine control over the subscription lifecycle (initialize, finalize)

Unfortunately, this would create an undesired dependency on user credentials from the controller.

Surfacing those actions through the source adapter is not an option because frequent requests from the controller would prevent scaling the adapter to zero.

One solution would be to request a separate, less privileged set of credentials (`GetSubscriptionAttributes` permission) from the user, but this could lead to a tedious setup experience.

Another solution would be to rely on an ephemeral delegate GRPC service to create, query and finalize SNS subscriptions on behalf of the controller, but this is not much different from disabling scale-to-zero on the adapter.

PR #158, opened 2020-06-12
Refactor SNS source

antoineco:
Controller:
* switch source deployment mode from Deployment to Knative Service
* update all tests and internal libraries to handle Knative Services as well as Deployments

Adapter:
* subscribe to the adapter's public URL
* allow configuration of subscription attributes
* pass unmodified notification payload in CloudEvents
* handle termination

#157 is causing serious limitations to this source unfortunately.


pkg/reconciler/awssnssource/adapter.go
```diff
@@ -34,27 +34,49 @@ import (
...
+		var subsAttrsStr string
+		if subsAttrsJSON, err := json.Marshal(src.Spec.SubscriptionAttributes); err == nil {
+			subsAttrsStr = string(subsAttrsJSON)
+		}
+
+		/* FIXME(antoineco): we wouldn't need to know the service URL in
+		   advance if we could reconcile SNS subscriptions from the controller.
+		   Ref. https://github.com/triggermesh/aws-event-sources/issues/157
+		*/
+		adapterURL := fmt.Sprintf("%s://%s.%s.%s", cfg.Scheme, name, src.Namespace, cfg.Domain)
```
antoineco, line 77: This is a terrible thing to do. Unfortunately Pods can not introspect their own public URL. We wouldn't have to resort to such hack if we had a solution for #157.

pkg/reconciler/awssnssource/adapter.go
```diff
@@ -75,13 +97,18 @@ func adapterDeploymentBuilder(src *v1alpha1.AWSSNSSource, cfg *adapterConfig) co
...
 			resource.EnvVarFromSecret(common.EnvAccessKeyID,
 				src.Spec.Credentials.AccessKeyID.ValueFromSecret.Name,
 				src.Spec.Credentials.AccessKeyID.ValueFromSecret.Key),
 			resource.EnvVarFromSecret(common.EnvSecretAccessKey,
 				src.Spec.Credentials.SecretAccessKey.ValueFromSecret.Name,
 				src.Spec.Credentials.SecretAccessKey.ValueFromSecret.Key),
 			resource.EnvVars(cfg.configs.ToEnvVars()...),
+			/* FIXME(antoineco): default metrics port 9090 overlaps with queue-proxy
+			 */
+			resource.EnvVar(source.EnvMetricsCfg, ""),
```
antoineco, line 111: This is being tackled in knative/eventing#3316

pkg/reconciler/testing/reconcile.go
```diff
@@ -231,6 +232,24 @@ func assertPopulatedSource(t *testing.T, src v1alpha1.AWSEventSource) {
...
+	var kind, resource string
+
+	switch object.(type) {
+	case *appsv1.Deployment:
+		kind = "Deployment"
+		resource = "deployments"
+	case *servingv1.Service:
+		kind = "Service"
+		resource = "services"
+	}
```
antoineco, line 248: I'm open to suggestions to do that in a more generic manner.

pkg/reconciler/testing/reconcile.go
```diff
@@ -294,18 +313,39 @@ func withoutSink(src v1alpha1.AWSEventSource) {
...
-	src.GetStatus().PropagateAvailability(nil)
+func unknownDeployedWithError(adapter runtime.Object) sourceOption {
+	var nilObj runtime.Object
+
+	switch adapter.(type) {
+	case *appsv1.Deployment:
+		nilObj = (*appsv1.Deployment)(nil)
+	case *servingv1.Service:
+		nilObj = (*servingv1.Service)(nil)
+	}
```
antoineco, line 344: Here as well. We could do it better if we could somehow access the object's GroupVersionKind. Suggestions welcome.

pkg/reconciler/testing/reconcile.go
odacremolbap: I don't think there is an universal solution.
```go
metaObj.GetObjectKind()
```
could be use for kinds, but then you would need the _downcase-and-pluralize_ for the resource. Imho think it looks simpler as you wrote it.

pkg/apis/sources/v1alpha1/common_lifecycle.go
```diff
@@ -83,3 +98,29 @@ func (s *AWSEventSourceStatus) PropagateAvailability(d *appsv1.Deployment) {
 
 	awsEventSourceConditionSet.Manage(s).MarkFalse(ConditionDeployed, ReasonUnavailable, msg)
 }
+
+func (s *AWSEventSourceStatus) propagateServiceAvailability(ksvc *servingv1.Service) {
+	if ksvc == nil {
+		awsEventSourceConditionSet.Manage(s).MarkUnknown(ConditionDeployed, ReasonUnavailable,
+			"The status of the adapter Service can not be determined")
+		return
```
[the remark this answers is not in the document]
antoineco: Yes that's a valid point, also I may want to reconsider setting things to "Unknown".

My reasoning was: what if the target was healthy, but then for a few cycles the k8s API returns errors? Do we wipe the URL and force triggers - which depend on the URL - to fail? Or do we admit we have no clue in what state the target is and decide not to touch the status?

The nil case is an edge case, and that's why I remained a bit conservative with it. I don't want transitive API errors to break the event flow.

Opinions welcome!

pkg/reconciler/testing/reconcile.go
antoineco: Even this doesn't work because API responses don't set TypeMeta, all you get is a typed object and you must use that to determine the API version and kind.

I always found that design weird, and there is a forever old k8s issue about it.

pkg/apis/sources/v1alpha1/common_lifecycle.go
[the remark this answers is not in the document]
odacremolbap: makes sense, setting to `Unknown` might mean `don't trust me a lot at this moment`.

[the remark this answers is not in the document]
antoineco: Maybe one way to unset the URL in a more conservative manner would be to consume the "last transition" date-time, compare to `time.Now()`, and unset if the diff is > x minutes.

PR #159, opened 2020-06-15
AWS CodeCommit source - avoid "for" loops

tzununbekov:
- `preparePullRequests` doesn't have `for` loop anymore:
  it's not explicitly documented, but it looks like `nextToken` in response is set only when we either specify list limit or there are more then 1000 items in list. However, our current loop break [condition](https://github.com/triggermesh/aws-event-sources/blob/b39f3d67459a8b0e6fdae3f5e5d773211df0ba47/pkg/adapter/awscodecommitsource/adapter.go#L241) doesn't seem to do a [correct](https://github.com/awsdocs/aws-doc-sdk-examples/blob/master/go/example_code/cloudwatch/CloudWatchGetLogEvents.go#L66) check anyway and we're not hanging there forever only because `nextToken` is not used.
- `sendPushEvent` and `sendPREvent` functions merged into a single `sendEvent` with event type assertion

_mentioned in #58_

### DIFF
--- a/pkg/adapter/awscodecommitsource/adapter.go
+++ b/pkg/adapter/awscodecommitsource/adapter.go
@@ -164,7 +164,7 @@ func (a *adapter) Start(stopCh <-chan struct{}) error {
 
 			for _, pr := range pullRequests {
 				resetBackoff = true
-				err = a.sendPREvent(pr)
+				err = a.sendEvent(pr)
 				if err != nil {
 					a.logger.Errorw("Failed to send PR event", "error", err)
 					return resetBackoff, nil
@@ -201,7 +201,7 @@ func (a *adapter) processCommits() error {
 
 	lastCommit = *commitOutput.Commit.CommitId
 
-	err = a.sendPushEvent(commitOutput.Commit)
+	err = a.sendEvent(commitOutput.Commit)
 	if err != nil {
 		return fmt.Errorf("failed to send push event: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


pkg/adapter/awscodecommitsource/adapter.go
```diff
@@ -216,68 +216,49 @@ func (a *adapter) preparePullRequests() ([]*codecommit.PullRequest, error) {
...
-	if err := event.SetData(cloudevents.ApplicationJSON, pullRequest); err != nil {
-		return fmt.Errorf("failed to set event data: %w", err)
+
+	switch v := codeCommitEvent.(type) {
+	case *codecommit.PullRequest:
+		event.SetType(v1alpha1.AWSEventType(a.arn.Service, prEventType))
+		err := event.SetData(cloudevents.ApplicationJSON, v)
+		if err != nil {
+			return fmt.Errorf("failed to set event data: %w", err)
+		}
```
antoineco, line 253: Does it matter that the interface is asserted here? Can't we move that outside of the switch? `SetData` accepts interfaces.

pkg/adapter/awscodecommitsource/adapter.go
```diff
@@ -216,68 +216,49 @@ func (a *adapter) preparePullRequests() ([]*codecommit.PullRequest, error) {
...
-		prIDs = append(prIDs, pullRequestsOutput.PullRequestIds...)
-
-		for _, id := range prIDs {
-			pri := codecommit.GetPullRequestInput{PullRequestId: id}
-
-			prInfo, err := a.ccClient.GetPullRequest(&pri)
-			if err != nil {
-				return pullRequests, fmt.Errorf("failed to get PR info: %w", err)
-			}
+	//Get pull request IDs
```
antoineco, line 219: A convention more than a written rule, but it's idiomatic to have a leading space for comments that are not meant to be consumed by machines (https://github.com/golang/go/issues/37974#issuecomment-607510041).

PR #161, opened 2020-06-17
Update Knative dependencies to v0.15

antoineco:
No API changes as far as those event sources are concerned.


sebgoa:
will the sources still run on v0.14 knative installation ?

antoineco:
Yes, we only internal libraries here, we don't interact with Knative at all.

PR #164, opened 2020-07-02
Offload ARN parsing to JSON unmarshaler

antoineco:
Handling ARNs as plain strings requires parsing in multiple places (reconciliation, tests, ...), which feels quite unnecessary.

Because Kubernetes clients/informers automatically deserialize JSON objects to typed Go structs, we can use typed ARNs in our CRs and implement a custom (de-)serializer so that the JSON/YAML surface doesn't change for users.


pkg/apis/arn.go
```diff
@@ -0,0 +1,57 @@
...
+
+	*a = ARN(arn)
+
+	return nil
+}
+
+// MarshalText implements encoding.TextMarshaler.
+func (a ARN) MarshalText() ([]byte, error) {
+	return []byte(a.String()), nil
+}
```
antoineco, line 57: These methods are called automatically when Kubernetes clients perform a JSON (de-)serialization using `json.(Un)Marshal()` so that an ARN string can transparently be handled as a `ARN` object by our reconcilers.

pkg/apis/sources/v1alpha1/awscodecommit_lifecycle.go
```diff
@@ -37,7 +39,7 @@ func (s *AWSCodeCommitSource) GetSink() *duckv1.Destination {
 }
 
 // GetARN implements AWSEventSource.
-func (s *AWSCodeCommitSource) GetARN() string {
+func (s *AWSCodeCommitSource) GetARN() apis.ARN {
 	return s.Spec.ARN
 }
```
antoineco, line 44: `GetARN()` returns the typed ARN instead of a string, so we can now remove all calls to `arn.Parse()` from the reconciliation code.

pkg/reconciler/common/reconcile.go
```diff
@@ -46,21 +44,15 @@ var knativeServingAnnotations = []string{
 }
 
 // AdapterDeploymentBuilderFunc builds a Deployment object for a source's adapter.
-type AdapterDeploymentBuilderFunc func(arn arn.ARN, sinkURI *apis.URL) *appsv1.Deployment
+type AdapterDeploymentBuilderFunc func(sinkURI *apis.URL) *appsv1.Deployment
```
antoineco, line 47: Incidentally, we can also simplify the signature of adapter builder funcs.

PR #167, opened 2020-07-03
chart: adds icon

sameersbn:


antoineco:
We should consider changing that lambda logo :)

PR #178, opened 2020-07-28
housekeeping for v0.4.2 release

sameersbn:


cmd/awscodecommitsource/Dockerfile
```diff
@@ -20,7 +20,7 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal
 
 LABEL name "Triggermesh AWS CodeCommit Event Source"
 LABEL vendor "Triggermesh"
-LABEL version "v0.4.1"
+LABEL version="v0.4.2"
```
antoineco, line 23: Just for my own curiosity, why does this field need the addition of an equal sign?

sameersbn: it does not need the equal sign. it was the result of a copy pasta recipe

PR #183, opened 2020-08-04
Align AWSEventSource interface with other event sources repos

antoineco:
The exact same change set as in triggermesh/knative-sources#68, with the following addition specific to AWS sources: the `GetARN()` method is now replaced by its equivalent `AsEventSource()`, which is more generic.

See commit history for each individual step.


antoineco:
Thanks for the review @cab105! :raised_hands:

PR #186, opened 2020-08-10
Find adapter by label and owner

antoineco:
Implements a method that finds the receive adapter for a given source by label and owner, instead of making assumptions on the adapter's name.

Relates to triggermesh/knative-sources#59

### DIFF
--- a/pkg/reconciler/awscodecommitsource/adapter.go
+++ b/pkg/reconciler/awscodecommitsource/adapter.go
@@ -17,7 +17,6 @@ limitations under the License.
 package awscodecommitsource
 
 import (
-	"fmt"
 	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -30,8 +29,6 @@ import (
 	"github.com/triggermesh/aws-event-sources/pkg/reconciler/common"
 	"github.com/triggermesh/aws-event-sources/pkg/reconciler/common/resource"
 )
-
-const adapterName = "awscodecommitsource"
 
 const (
 	envBranch     = "BRANCH"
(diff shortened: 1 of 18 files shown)


pkg/reconciler/common/reconcile.go
```diff
@@ -253,3 +273,59 @@ func (r *GenericServiceReconciler) syncAdapterService(ctx context.Context,
...
+		}
+
+		gr = servingv1.Resource("service")
+	}
+
+	for _, obj := range objs {
+		if metav1.IsControlledBy(obj, src) {
+			return obj, nil
+		}
+	}
```
antoineco, line 319: This PR doesn't cover the edge case where multiple objects are owned by the source. Even if I wanted to, I wouldn't know what to do with the extraneous object(s): keep, delete, ...

pkg/reconciler/awscodecommitsource/adapter.go
```diff
@@ -50,8 +47,10 @@ type adapterConfig struct {
 // adapterDeploymentBuilder returns an AdapterDeploymentBuilderFunc for the
 // given source object and adapter config.
 func adapterDeploymentBuilder(src *v1alpha1.AWSCodeCommitSource, cfg *adapterConfig) common.AdapterDeploymentBuilderFunc {
+	adapterName := common.AdapterName(src)
```
antoineco, line 50: This change feels like a stretch to this PR, but because the `FindAdapter` method uses this new `AdapterName()` method under the hood, I figured it would make sense to also use it the adapter builder funcs right away, in a separate commit.

cab105: So this is to ensure that the adapterName conforms to knative's expectations going forward?

antoineco: Knative doesn't expect any particular format, I just needed a way to derive the "app" name from the Kubernetes object so I don't have to pass it as an argument to every function call.
Also we don't have to hard code some arbitrary name anymore, it's always gonna be the object type lower cased (which is what we used until now anyway).

Issue #185, opened 2020-08-07
SNS: avoid re-subscribing in every loop


antoineco:
Check whether the subscription exists before calling `Subscribe()` to avoid waking the receive adapter up unnecessarily on every reconciliation period.

PR #187, opened 2020-08-10
Reconcile SNS subscriptions outside of the receive adapter

### LINKED ISSUE #157: Reconcile SNS subcriptions outside of the receive adapter

antoineco:
Moves the reconciliation of the SNS subscription out of the receive adapter. This has several benefits:

1. The receive adapter doesn't need to be kept "warn" by sending a request to itself every 10 seconds. This was super hacky to start with.
1. The SNS subscription can now be cleanly finalized (deleted from the topic) when the corresponding source object is deleted from the cluster.
1. Status reporting (yay!):

    ```console
    $ kubectl get awssnssource/sample
    NAME     READY   REASON                 URL                                              SINK                                        AGE
    sample   False   SubscriptionRejected   http://awssnssource-sample.default.example.com   http://sockeye.default.svc.cluster.local/   6m
    ```

    ```console
    $ kubectl describe awssnssource/sample
    ...
    Events:
      Type     Reason           Age                    From                     Message
      ----     ------           ----                   ----                     -------
      Warning  FailedSubscribe  4m27s (x3 over 5m37s)  awssnssource-controller  Error subscribing endpoint "http://awssnssource-sample.default.example.com" to SNS topic "arn:aws:sns:eu-central-1:043455440429:triggermeshtest": InvalidParameter: Invalid parameter: Unreachable Endpoint
    ```

One drawback: the controller now needs `GET` permissions on Kubernetes Secrets.

Out-of-scope, to implement in a follow up task: #185

Closes #157


antoineco:
@sameersbn there are a few changes that require your attention, all the changes to Helm charts especially, which I performed manually.

pkg/reconciler/awssnssource/subscription.go
```diff
@@ -0,0 +1,263 @@
...
+
+	// abandon if the subscription's ARN was never written to the source's status
+	if subsARN == nil {
+		return nil
+	}
+
+	spec := src.(apis.HasSpec).GetUntypedSpec().(v1alpha1.AWSSNSSourceSpec)
+
+	snsClient, err := newSNSClient(r.secretsCli(src.GetNamespace()), spec.ARN.Region, &spec.Credentials)
+	if err != nil {
```
antoineco, line 124: In case the user deletes the secret containing the AWS security credentials _before_ deleting the source itself, the finalizer will fail forever on the error
```
secrets "awscreds" not found
```
preventing the object from being deleted.

We need to check the returned error here, and give up if that error is unrecoverable.

antoineco: Done

PR #198, opened 2020-09-08
Implement json.Marshaler instead of encoding.TextMarshaler

antoineco:
This is required in order to be able to use a dynamic client to interact with those types, because `Unstructured` converters only support the JSON marshaler.

Extracted from #196

### DIFF
--- a/pkg/apis/arn.go
+++ b/pkg/apis/arn.go
@@ -17,7 +17,7 @@ limitations under the License.
 package apis
 
 import (
-	"encoding"
+	"encoding/json"
 	"fmt"
 
 	"github.com/aws/aws-sdk-go/aws/arn"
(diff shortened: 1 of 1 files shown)


pkg/apis/arn.go
```diff
@@ -28,19 +28,23 @@ import (
...
 
-// UnmarshalText implements encoding.TextUnmarshaler.
-func (a *ARN) UnmarshalText(data []byte) error {
-	dataStr := string(data)
+// UnmarshalJSON implements json.Unmarshaler.
+func (a *ARN) UnmarshalJSON(data []byte) error {
+	var dataStr string
+	if err := json.Unmarshal(data, &dataStr); err != nil {
+		return err
+	}
```
antoineco, line 46: If the data passed to that method is a valid JSON string, it will look like `"arn:aws:..."` (quotes included). Passing those bytes through `json.Unmarshal` is safer than trying to manually trim the surrounding double-quotes.

antoineco (commented):
Notes to reviewers.

antoineco:
Merging because it's currently preventing me from moving on with e2e tests.

PR #199, opened 2020-09-08
cloudbuild: add (new) kaniko flags to speed up cloudbuilds

sameersbn:
The latest kaniko release has two new flags that boasts of upto 75% performance optimization


antoineco (approved):
Good news!

PR #210, opened 2020-10-26
circleci: update orb versions

sameersbn:
use the latest available version of the circle ci orbs. 
the latest version of the circleci/go org fixes go module caching during builds

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,9 @@
 version: 2.1
 
 orbs:
-  go: circleci/go@1.1.1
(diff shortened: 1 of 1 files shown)


antoineco (approved):

sameersbn:
@antoineco this change will use the latest versions of those orbs. but more importantly it appears there are changes in the go orb after which the go module cache/restore actually works

antoineco:
Ahhhh great news! Support for caching has been pretty terrible so far so I have high hopes 😄

PR #214, opened 2020-10-30
Refactor Backoff to allows min below 1s

antoineco:
Also refactor tests to provide more control over the test logic

### DIFF
--- a/pkg/adapter/common/backoff.go
+++ b/pkg/adapter/common/backoff.go
@@ -26,31 +26,32 @@ import (
 const (
 	expFactor = 2
 
-	minBackoff = 1 * time.Second
-	maxBackoff = 32 * time.Second
+	defaultMinBackoff = 1 * time.Second
+	defaultMaxBackoff = 32 * time.Second
 )
 
-// Backoff provides a simple exponential backoff mechanism
+// Backoff provides a simple exponential backoff mechanism.
 type Backoff struct {
 	step     int
 	factor   float64
 	min, max time.Duration
 }
 
-// RunFunc is a user function that polls data from a source and sends it as
-// a cloudevent to a sink.
-// RunFunc must return (bool, error) values where bool is true if poll backoff duration
-// must be reset and error is the result of function execution.
-type RunFunc func(context.Context) (bool, error)
+// RunFunc is a user function that polls data from a source and sends it as a
+// CloudEvent to a sink.
+// RunFunc must return (bool, error) values where bool is true if the poll
+// backoff duration must be reset, and error is the result of the function's
+// execution.
+type RunFunc func(context.Context) (bool /*reset*/, error /*exit*/)
 
-// NewBackoff accepts optional values for minimum and maximum wait period
-// and return new instance of Backoff structure
+// NewBackoff accepts optional values for minimum and maximum wait period and
+// returns a new instance of Backoff.
 func NewBackoff(args ...time.Duration) *Backoff {
 	backoff := &Backoff{
 		step:   0,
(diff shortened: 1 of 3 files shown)


pkg/adapter/common/backoff_test.go
```diff
@@ -109,15 +108,16 @@ func TestDuration(t *testing.T) {
...
 		fn         func(context.Context) (bool, error)
 		waitReturn bool
 		err        error
 	}{
-		"force stop with no errs": {
+		"force stop with no err": {
+			fn: func(ctx context.Context) (bool, error) {
+				// never terminates by itself
+				return true, nil
+			},
```
antoineco, line 120: I encountered a race when `fn` was `nil` and the `doneCh` wasn't closed fast enough.

pkg/adapter/common/backoff.go
```diff
@@ -68,9 +69,9 @@ func NewBackoff(args ...time.Duration) *Backoff {
 	return backoff
 }
 
-// Duration can be used to get exponential backoff duration calculated for each new step
+// Duration returns the exponential backoff duration calculated for the current step.
 func (b *Backoff) Duration() time.Duration {
-	dur := time.Duration(float64(b.min)*math.Pow(b.factor, float64(b.step)) - float64(1*time.Second))
+	dur := time.Duration(float64(b.min)*math.Pow(b.factor, float64(b.step)) - float64(b.min))
```
antoineco, line 74: This change is what makes it work for min backoff values below 1s.

pkg/adapter/common/backoff_test.go
```diff
@@ -127,47 +127,90 @@ func TestRun(t *testing.T) {
 			waitReturn: true,
 			err:        assert.AnError,
 		},
-		"fn resets backoff duration": {
-			fn: func(ctx context.Context) (bool, error) {
-				counter++
-				if counter > 3 {
-					if time.Since(start).Round(time.Second) != 3*time.Second {
```
antoineco, line 134: This time-based expectation was very optimistic, considering `start` is shared between all tests, and delaying one previous test could have  made this one fail.

I rewrote it using a channel to control the number of executions of the function from the main test logic.

antoineco (commented):
Notes

pkg/adapter/common/backoff_test.go
tzununbekov: nice, I remember I was having a hard time with this test and the time-based check was a sort of desperate decision

sebgoa (approved):
nothing is standing out after a quick check

pkg/adapter/common/backoff_test.go
```diff
@@ -127,47 +128,89 @@ func TestRun(t *testing.T) {
...
+		// of the main test logic.
+		latch := make(chan struct{})
+
+		fn := func(ctx context.Context) (bool, error) {
+			latch <- struct{}{}
+
+			defer func() {
+				runCounter++
+				latch <- struct{}{}
+			}()
```
tzununbekov, line 181: I'm not sure that I completely understand why do we need that deferred function with the counter, but the thing with double locking latches is something that I want to use somewhere else :slightly_smiling_face:

tzununbekov (approved):

pkg/adapter/common/backoff_test.go
antoineco: Nice catch, this is a leftover from an older version of this code! We only need to defer the "unlatch" here.

antoineco: Removed

PR #219, opened 2020-11-13
DynamoDB: Process shards in parallel, recheck stream

### LINKED ISSUE #218: DynamoDB source doesn't generate any event

antoineco:
Fixes #218 

This is a complete rewrite of the DynamoDB source, so beware that the PR review might be quite technical.

e2e tests written by @sameersbn uncovered serious issues with the DynamoDB source. In most cases, it doesn't process any event because it doesn't iterate over shards correctly.

The "correct" way to iterate over shards is to read the _next shard_ from the response to `GetRecords` API calls. `GetShardIterator` only returns the _initial_ iterator, but the records are not necessarily accessible at this particular iterator.

There is another quirk: the chain of `NextShardIterator`s returned by `GetRecords` is endless. The iterator only becomes `nil` when the shard gets sealed (marked as read only), which also happens regularly and automatically. Therefore, we need to run those records processors as background routines.

The documentation at https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.html is limited but explains that shards are _ephemeral_  and can _split into multiple new shards_ automatically, so we also need to recheck the stream from time to time (every 15s in my code), which we never did before.

This implementation is inspired by the example at https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.LowLevel.Walkthrough.html, with all the concurrency added on top.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ replace (
 )
 
 require (
-	github.com/aws/aws-sdk-go v1.31.12
+	github.com/aws/aws-sdk-go v1.35.28
 	github.com/cloudevents/sdk-go/v2 v2.2.0
 	github.com/google/go-cmp v0.5.2
 	github.com/kelseyhightower/envconfig v1.4.0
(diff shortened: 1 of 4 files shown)


antoineco:
I still need to rewrite tests as well, so please ignore the CI for now.

antoineco:
Tests added:
```
--- PASS: TestAdapter (0.02s)
PASS
coverage: 73.6% of statements
ok      github.com/triggermesh/aws-event-sources/pkg/adapter/awsdynamodbsource  0.185s  coverage: 73.6% of statements
```

sameersbn (approved):
as highlighted by the tests, there are some unused variables. other than that lgtm

PR #223, opened 2020-11-25
sqs: Set resources and limits

antoineco:
Based on observations from a load test with 100,000 messages, with 1 then 10 replicas (1 worker per replica for the time being).

image

---

I'm dropping the corresponding throughput profile here as a bonus (~40 msg/s _per instance_ when the queue and receiver are both located in the U.S.):

result-thrpt


sebgoa (approved):
nice to see something predictable :)

cab105 (approved):

PR #225, opened 2020-11-30
sqs: Optimize for concurrency

### LINKED ISSUE #222: Profile SQS source
The goal is to find the best trade-off between CPU shares and number of concurrent message receivers, so we can set a sane default for:
* the container's `resources.limits.cpu`
* the number of goroutines running [`ReceiveMessage`](https://docs.aws.amazon.com/sdk-for-go/api/service/sqs/#SQS.ReceiveMessage) concurrently

antoineco:
Rewrite of the source adapter to spawn concurrent message processors instead of executing a single loop sequentially.

The number of receivers and senders is based on the number of available CPU cores.

Closes #222

### DIFF
--- a/cmd/awssqssource/main.go
+++ b/cmd/awssqssource/main.go
@@ -17,11 +17,30 @@ limitations under the License.
 package main
 
 import (
+	"runtime"
+
 	"knative.dev/eventing/pkg/adapter/v2"
 
 	"github.com/triggermesh/aws-event-sources/pkg/adapter/awssqssource"
 )
 
 func main() {
+	setMaxProcs(runtime.NumCPU())
+
 	adapter.Main("awssqssource", awssqssource.NewEnvConfig, awssqssource.NewAdapter)
+}
+
+// setMaxProcs sets the number of threads that can be used by the current
+// process.
+//
+// Knative uses uber-go/automaxprocs to automatically determine the number of
+// threads (GOMAXPROCS) available to the process based on CPU quotas (e.g.
+// 'cpu.request <= 1' translates to 1 thread, regardless of the number of
+// physical cores).
+// This event source has a very predictable CPU profile: it spends most of its
+// time sending network requests without performing any computation on the
+// results, so we assume the defined CPU limit allows all CPU cores to be used
+// without starving on CPU time.
+func setMaxProcs(procs int) int {
+	return runtime.GOMAXPROCS(procs)
 }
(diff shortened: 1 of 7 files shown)


antoineco:
## Post-optimization analysis

Recap:
* CPU quota set at **200m**
* Multiple SQS message receivers (**1 per CPU core**)
* Multiple CE senders (**1 per CPU core**)
* 1 message deleter: blocks every 10 messages to delete processed messages in batch (10 is the max size) but is twice as fast as receivers, which should be an acceptable trade-off

### Overview: resources usage and throughput

Although the CPU is now being throttled due to CPU quotas, we can see that this throttling remains constantly below 20%:

overview-cpu

The memory usage is still stable and within the defined request for the entire duration of the load test:

overview-mem

The throughput has improved decently, from ~40 messages/sec (without CPU quotas) to ~110 messages/sec (with CPU quotas). 

result-thrpt

### Scheduling

Overall threads are now better utilized, especially with receiving, sending and deleting happening concurrently across 2 threads. However, we can observe that goroutines spend most of their time synchronizing between each other.

Garbage collection takes also longer due to CPU throttling.

pprof-overview

_receivers_

pprof-receivers

_senders_

pprof-senders

_deleter_

pprof-deleter

### Wrap up

I could perform another test without any quota, but I'm not convinced it would improve performances drastically. Most of the time is spent syncing between goroutines, which is probably where we have to focus next.

With 2 senders and only one deleter, the deleter is very likely to be a bottleneck. Especially if the flow of messages is constant, the deletion buffer will keep filling no matter its size if we don't parallelize that work better.

sebgoa:
so this is one Pod with a 200m CPU quota ?

meaning we could run 5 of those on one core, do I get that right ?

antoineco:
Yes and no. The share applies across all cores, it translates to a fraction of CPU time based on a 100ms period. This code is written to leverage multiple cores _no matter the CPU quota_. The number depends on the actual CPU type, detected at runtime). A limit below "1" doesn't mean that the Go runtime will only see 1 core and schedule everything on 1 core.

I haven't been able to figure how exactly the kernel applies the share across multiple cores though, because if you look at the graph, the "CPU usage" isn't capped at the limit (yellow line), yet some throttling is applied.

I want to spend some time to day to tweak that and compare what really happens when no limit is defined.

sebgoa:
ok got it.

In general terms as you parallelize this you will start hitting a "limit" where sequential routines will take over the bulk of the time spent. It will become less and less easy to make gains.

antoineco:
The same load test without CPU limits yields the same performance:

result-thrpt

sebgoa:
sorry,, I clicked on merge instead of approve :(

PR #239, opened 2020-12-17
Address my comments in PR 238 (CloudWatch)

antoineco:
@cab105 this should address most of my comments.

I recommend checking "Hide whitespace changes" to review the CRD changes.

image


pkg/adapter/awscloudwatchlogsource/adapter.go
```diff
@@ -80,33 +74,31 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClie
...
 
 	env := envAcc.(*envConfig)
-	a := common.MustParseARN(env.ARN)
 
-	awsCfg := aws.NewConfig()
-	awsCfg.WithCredentials(credentials.NewStaticCredentials(env.AccessKey, env.SecretKey, ""))
-	cfg := session.Must(session.NewSession(awsCfg.
-		WithRegion(a.Region).
-		WithMaxRetries(5),
+	arn := common.MustParseARN(env.ARN)
```
cab105, line 78: This is the one area where I've been given warning about the package name coinciding with the variable name, which is why I stuck with `a` instead of `arn`.

pkg/adapter/awscloudwatchlogsource/adapter.go
```diff
@@ -135,38 +127,19 @@ func (a *adapter) Start(ctx context.Context) error {
...
-			}
+
+	// Wake up every pollingInterval, and retrieve the logs
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+
+		case t := <-poll.C:
+			a.CollectLogs(t)
```
cab105, line 139: In this case, I opted to use the go-routine to ensure log and metric collection/event generation did not interfere with the polling interval.

antoineco: The ticker keeps running even if a function is blocking the select, unlike a timer. Is that what you're worried about?

pkg/adapter/awscloudwatchlogsource/adapter.go
antoineco: Where do you see this warning?

antoineco: Reverted in cf58977f173705371da6b364f3c2cb6292ca4c6e

cab105: Feature of the IDE I'm using: https://www.jetbrains.com/help/go/code-inspections-in-go.html#Probable_bugs


Imported package name as name identifier | Reports declarations of variables, arguments or functions that overlap with the used import.                     While  legal, such declarations will make using the package exported  identifiers impossible after the declaration or create confusion when  reading the code. | Warning
-- | -- | --

pkg/adapter/awscloudwatchlogsource/adapter.go
cab105: More like:
1. Ticker fires every 2 minutes
2. Processing takes longer than two minutes (suppose an extra 30 seconds)
3. On next loop, processing uses the current time, and the starting time interval from now - 2 minutes resulting in a loss of metric/log data of that delta.

Granted this could be worked around by passing the previous fire time with the current time into the collection functions, but this also raises a concern of a chain reaction delay where the processing of the first event delays the processing of the second event, and so on with the delays growing over time. The goroutine approach fans out the processing so the delays won't impact the primary server thread unless they grow in enough number to slow the overall system down and summon the thundering herd.

cab105:
Thank you again for your help! Will incorporate this with my other cleanup work. Meanwhile will move my other PR into a draft state, and open it up again once I have it cleaned up a little.

pkg/adapter/awscloudwatchlogsource/adapter.go
antoineco: The question here is: is it reasonable to run multiple `CollectLogs` in parallel? I'm reverting this change, but the design looks shaky to me, especially with this calculation based on the interval. If you don't base your startTime on the timestamp of the last collected log, a delayed tick will potentially cause missed entries. But that's off topic here.

PR #241, opened 2020-12-22
sns: Remove unnecessary context and WaitGroup

### LINKED ISSUE #240: sns: Occasional test failure due to context cancelation
https://github.com/triggermesh/aws-event-sources/blob/27cce63bb0c09a5a0eb377718d2f3af12f6becf7/pkg/adapter/awssnssource/adapter.go#L137

At that line, `ctx` has already been canceled so the error below is returned

```
Failed
=== RUN   TestStart
    logger.go:130: 2020-12-21T23:42:16.148Z	INFO	awssnssource/adapter.go:105	Shutdown signal received. Terminating
    logger.go:130: 2020-12-21T23:42:16.148Z	INFO	awssnssource/adapter.go:120	Serving on port 8080
    logger.go:130: 2020-12-21T23:42:16.148Z	INFO	awssnssource/adapter.go:135	Shutting server down
    adapter_test.go:76: Receiver returned an error: error during server shutdown: context canceled
--- FAIL: TestStart (0.00s)
```

`shutdownCtx` should be based on an empty context instead.

antoineco:
Fixes #240

### DIFF
--- a/pkg/adapter/awssnssource/adapter.go
+++ b/pkg/adapter/awssnssource/adapter.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"sync"
 	"time"
 
 	"go.uber.org/zap"
@@ -95,17 +94,6 @@ const (
 
 // Start implements adapter.Adapter.
 func (a *adapter) Start(ctx context.Context) error {
-	// ctx gets canceled to stop goroutines
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
-
-	// handle stop signals
-	go func() {
-		<-ctx.Done()
-		a.logger.Info("Shutdown signal received. Terminating")
-		cancel()
-	}()
-
 	http.HandleFunc("/", a.handleNotification)
 	http.HandleFunc("/health", healthCheckHandler)
 
@@ -113,13 +101,9 @@ func (a *adapter) Start(ctx context.Context) error {
 	serverErrCh := make(chan error)
 	defer close(serverErrCh)
 
-	var wg sync.WaitGroup
-
-	wg.Add(1)
 	go func() {
 		a.logger.Info("Serving on port " + serverPort)
 		serverErrCh <- server.ListenAndServe()
-		wg.Done()
 	}()
 
 	var err error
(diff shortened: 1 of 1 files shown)


pkg/adapter/awssnssource/adapter.go
```diff
@@ -95,31 +94,16 @@ const (
...
 func (a *adapter) Start(ctx context.Context) error {
-	// ctx gets canceled to stop goroutines
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
-
-	// handle stop signals
-	go func() {
-		<-ctx.Done()
-		a.logger.Info("Shutdown signal received. Terminating")
-		cancel()
```
antoineco, line 106: This call to `cancel()` did absolutely nothing, since `ctx` is already "done" 2 lines above.

pkg/adapter/awssnssource/adapter.go
```diff
@@ -95,31 +94,16 @@ const (
...
 	http.HandleFunc("/", a.handleNotification)
 	http.HandleFunc("/health", healthCheckHandler)
 
 	server := &http.Server{Addr: ":" + serverPort}
 	serverErrCh := make(chan error)
 	defer close(serverErrCh)
 
-	var wg sync.WaitGroup
-
-	wg.Add(1)
```
antoineco, line 118: No need for a WaitGroup when there is a single goroutine to wait for and that goroutine already sends its result to a channel (`serverErrCh` in this case)

pkg/adapter/awssnssource/adapter.go
```diff
@@ -129,22 +113,20 @@ func (a *adapter) Start(ctx context.Context) error {
...
 		}
-		cancel()
 
 	case <-ctx.Done():
 		a.logger.Info("Shutting server down")
 
-		shutdownCtx, cancelTimeout := context.WithTimeout(ctx, serverShutdownGracePeriod)
-		defer cancelTimeout()
-		if shutdownErr := server.Shutdown(shutdownCtx); shutdownErr != nil {
+		ctx, cancel := context.WithTimeout(context.Background(), serverShutdownGracePeriod)
```
antoineco, line 120: As explained in the referenced issue, `ctx` has already expired at that point, so `WithTimeout()` needs to be based on a fresh context, otherwise it will return immediately instead of waiting for a graceful shutdown.

antoineco (commented):
Notes to reviewers

PR #253, opened 2021-01-15
Add JSON schemas for emitted event types

antoineco:
For unmodified exported type, the following program was used to generate schemas:

```go
package main

import (
        "encoding/json"
        "fmt"

        "github.com/alecthomas/jsonschema"
)

func main() {
        typ := // the type to generate a schema for

        schema := jsonschema.Reflect(typ)

        b, err := json.MarshalIndent(schema, "", "  ")
        if err != nil {
                panic(fmt.Errorf("marshaling schema: %w", err))
        }

        fmt.Print(string(b), "\n")
}
```

Others, like `com.amazon.sns.notification.json`, were written by hand.

### DIFF
--- a/schemas/com.amazon.cloudwatch.metrics.message.json
+++ b/schemas/com.amazon.cloudwatch.metrics.message.json
@@ -1,0 +1,22 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "$ref": "#/definitions/MessageData",
+  "definitions": {
+    "MessageData": {
+      "required": [
+        "Code",
+        "Value"
+      ],
+      "properties": {
+        "Code": {
+          "type": "string"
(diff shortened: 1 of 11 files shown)


sebgoa:
where are those schemas coming from ? Should we put a link to the source somewhere.

antoineco:
They come from us.

odacremolbap (approved):
I think all AWS fields are pointer types hence nil-able, but they are set as required at the schema.
I haven't checked all structures but the `required` fields at the schema sounds right at the ones I've looked at. 

+1 because they look right, I haven't tested it.

antoineco:
I'm sure some small adjustments will be necessary in the future, but this should be a good base we can iterate on once we try validating real events.

PR #254, opened 2021-01-18
Update Knative dependencies to v0.20.0

antoineco:
Routine dependency update post Knative v0.20.0 release.

Nothing to change in our code while updating to this release.

---

_Usual disclaimer: this is **not** an upgrade of Knative. Updating those dependencies does **not** break the compatibility with the deployed Knative version considering there is no change in public APIs._


sameersbn (approved):
lgtm

PR #262, opened 2021-01-28
Make the SNS source multi-tenant (part 1)

antoineco:
Embed a skeleton of reconciler into the AWSSNSSource adapter that implements both Reconciler (add HTTP route) and Finalizer (remove HTTP route).

At this stage, the adapter still handles a single endpoint at `/`, the actual routing logic is in TODO. I'm marking the PR as ready nevertheless, because I think the code will be easier to review if it is split into two logical PRs. (thoughts?)

### DIFF
--- a/pkg/adapter/awssnssource/adapter.go
+++ b/pkg/adapter/awssnssource/adapter.go
@@ -1,11 +1,11 @@
 /*
-Copyright (c) 2019-2020 TriggerMesh Inc.
+Copyright (c) 2019-2021 TriggerMesh Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-   http://www.apache.org/licenses/LICENSE-2.0
+    http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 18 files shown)


JeffNeff (approved):
This part was fairly easy to follow. There are some concepts that I have not seen before so it was a good read! Cant wait for pt two!.

pkg/adapter/awssnssource/controller.go
```diff
@@ -0,0 +1,55 @@
...
+		r := &Reconciler{
+			adapter: a.(MTAdapter),
+		}
+		impl := reconcilerv1alpha1.NewImpl(ctx, r, controller.Opts(component))
+
+		informerv1alpha1.Get(ctx).Informer().AddEventHandler(pkgcontroller.HandleAll(impl.Enqueue))
+
+		return impl
+	}
+}
```
antoineco, line 55: This code sums up this PR quite well I think.

In the shared main, Knative initializes the adapter as usual, but now it also passes that adapter to `NewController`. The adapter exposes methods that the reconciler calls during reconciliation. In other words, the reconciler is in charge of triggering a change in the adapter whenever a source object changes.

pkg/adapter/awssnssource/reconciler.go
```diff
@@ -0,0 +1,112 @@
...
+		// triggers a new reconciliation.
+		return controller.NewPermanentError(reconciler.NewEvent(corev1.EventTypeWarning, "NotReady",
+			"Source is not ready yet. Skipping adapter configuration"))
+	}
+
+	err := r.adapter.RegisterHandlerFor(ctx, src)
+	if err != nil {
+		return reconciler.NewEvent(corev1.EventTypeWarning, ReasonHandlerFailedRegister,
+			"Failed to register HTTP handler: %s", err)
+	}
```
antoineco, line 86: Here is an example of call from the reconciler to a method exposed by the adapter.

pkg/adapter/awssnssource/adapter.go
```diff
@@ -196,3 +218,20 @@ func healthCheckHandler(w http.ResponseWriter, _ *http.Request) {
...
+// DeregisterHandlerFor implements MTAdapter.
+func (a *adapter) DeregisterHandlerFor(ctx context.Context, src *v1alpha1.AWSSNSSource) error {
+	// TODO(antoineco): implement routing
+	return nil
+}
+
+// PropagateCondition implements MTAdapter.
+func (a *adapter) PropagateCondition(ctx context.Context, src *v1alpha1.AWSSNSSource, cond *apis.Condition) error {
+	return status.PropagateCondition(ctx, a.statusPatcher, src, cond)
+}
```
antoineco, line 237: As mentioned in the PR description, I left the implementation of the HTTP router in TODO and will tackle it in another PR, so this one can focus purely on the interactions between the reconciler and the adapter.

antoineco (commented):
Short walkthrough of this PR

PR #266, opened 2021-02-16
Fix bug in cloudwatch name setting

cab105:
Observed a bug while prepping a demo where the string pointer being used for the metric name is associated with an object that is reused in a for loop, and not the underlying string. This is resulting in the same pointer address being used, and as a result the last name defined within the metric being used for all entries.

Additionally, this PR performs a slight optimization to appending metric queries.

### DIFF
--- a/pkg/adapter/awscloudwatchsource/adapter.go
+++ b/pkg/adapter/awscloudwatchsource/adapter.go
@@ -112,21 +112,19 @@ func parseQueries(rawQuery string) ([]*cloudwatch.MetricDataQuery, error) {
 	}
 
 	for _, v := range rawQueries {
-		var q cloudwatch.MetricDataQuery
+		name := v.Name
 
 		if v.Expression != nil {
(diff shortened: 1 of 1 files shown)


antoineco:
The [`exportloopref`](https://github.com/kyoh86/exportloopref) linter reports the same issue at lines 137,138.

```console
$ golangci-lint run ./pkg/adapter/awscloudwatchsource --enable=exportloopref
pkg/adapter/awscloudwatchsource/adapter.go:137:12: exporting a pointer for the loop variable v (exportloopref)
                        Name:  &v.Name,
                                ^
pkg/adapter/awscloudwatchsource/adapter.go:138:12: exporting a pointer for the loop variable v (exportloopref)
                        Value: &v.Value,
                                ^
```

antoineco:
☝️  actually a false positive. You're taking the address of a string field, not of the loop variable, all good.

antoineco (approved):

PR #267, opened 2021-03-26
Add event source for Amazon S3

### LINKED ISSUE #162: Source for S3
We can get S3 bucket notifications from SQS if we turn on event notification for a bucket manually.

Maybe we can have an S3 source that does the config manually and re-uses the SQS source..

antoineco:
Closes #162

Implementation that receives event notifications from a S3 bucket [over a SQS queue](https://docs.aws.amazon.com/AmazonS3/latest/userguide/NotificationHowTo.html).

It does so by spawning an instance of `AWSSQSSource` which uses the same credentials as the ones passed to the `AWSS3Source`.

The user can set the `queueARN` spec attribute to send events to an existing queue. If this attribute is not set, a queue is created with default settings.


pkg/client/s3/client.go
```diff
@@ -0,0 +1,89 @@
...
+func (g *ClientGetterWithSecretGetter) Get(src *v1alpha1.AWSS3Source) (Client, SQSClient, error) {
+	creds, err := aws.Credentials(g.sg(src.Namespace), &src.Spec.Credentials)
+	if err != nil {
+		return nil, nil, fmt.Errorf("retrieving AWS security credentials: %w", err)
+	}
+
+	sess := session.Must(session.NewSession(awscore.NewConfig().
+		WithRegion(src.Spec.ARN.Region).
+		WithCredentials(credentials.NewStaticCredentialsFromCreds(*creds)),
+	))
```
odacremolbap, line 75: nit - personal preference:

session.Must seems to be doing
```go
func Must(object)
	if err != nil {
		panic(err)
	}
        return object
```

Since an error is already returned from this func, I would use it and let caller decide if to panic.

pkg/reconciler/awss3source/bucket.go
```diff
@@ -0,0 +1,254 @@
...
+
+	status := &typedSrc.Status
+
+	bucketARN := typedSrc.Spec.ARN
+
+	notifCfg, err := getNotificationsConfig(ctx, cli, bucketARN.Resource)
+	switch {
+	case isDenied(err):
+		status.MarkNotSubscribed(v1alpha1.AWSS3ReasonAPIError, "Access denied to S3 API")
+		return controller.NewPermanentError(reconciler.NewEvent(corev1.EventTypeWarning, ReasonFailedSubscribe,
```
odacremolbap, line 57: nit

not wrong but maybe also not right. If a user gets an error contacting the API, would the user fix the issue at AWS and expect that we reconcile automatically, or would they expect to re-create the source after they modify permissions?

odacremolbap (approved):
+1
3 opinions added, not blocking.

config/300-awss3.yaml
```diff
@@ -0,0 +1,213 @@
...
+                  - s3:ObjectRestore:*
+                  - s3:ObjectRestore:Post
+                  - s3:ObjectRestore:Completed
+                  - s3:ReducedRedundancyLostObject
+                  - s3:Replication:*
+                  - s3:Replication:OperationFailedReplication
+                  - s3:Replication:OperationNotTracked
+                  - s3:Replication:OperationMissedThreshold
+                  - s3:Replication:OperationReplicatedAfterThreshold
+              queueARN:
```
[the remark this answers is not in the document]
antoineco: Yes I added it as a fallback because S3 sends a test event with a different schema every time the configuration of event notifications changes in the bucket.

Now that you mention it, I could simply plan for that case and add a "testevent" type. The list of supported events is exhaustive in the spec after all.

pkg/client/s3/client.go
antoineco: I don't mind changing that, but then I'd do it in all sources in one go.

JeffNeff (approved):
wow.. beautiful code. great read. LGTM

pkg/adapter/awssqssource/adapter.go
```diff
@@ -93,6 +99,14 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClie
 
 	arn := common.MustParseARN(env.ARN)
 
+	var msgPrcsr MessageProcessor
+	switch env.MessageProcessor {
+	case "s3":
+		msgPrcsr = &s3MessageProcessor{ceSource: arn.String()}
+	default:
+		msgPrcsr = &defaultMessageProcessor{arn: arn}
+	}
```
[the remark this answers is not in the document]
antoineco: That's nicer 👍

pkg/reconciler/awss3source/bucket.go
antoineco: You're absolutely right. The approach I chose was a trade-off because I don't like error loops :)

There is a refresh configured in the informer for that reason. I have a personal preference for not letting reconcilers enter a loop if the credentials are obviously not authorized, but I get your point and fully agree.

I guess the right approach here would be something in between, like "retry, but only in 30s" or something like that.

The question here is: what do you think will happen more often? Credentials are not valid and need a refresh anyway, or credentials are missing a permission and need to change at AWS? (I don't have the answer)

config/300-awss3.yaml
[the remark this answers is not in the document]
odacremolbap: I wanted to understand that one, maybe `testevent` could be confusing. I think it is better as is for users that want to map to S3 events.

pkg/client/s3/client.go
odacremolbap: makes sense.

pkg/reconciler/awss3source/bucket.go
odacremolbap: yes, let's resolve this since no one is going to provide the answer.

@JeffNeff  mentioned this recently that he thought that producing error events from sources could be beneficial in some situations ... maybe permanent errors is one.

cab105 (approved):
Interesting approach.

pkg/adapter/awssqssource/adapter.go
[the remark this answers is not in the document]
antoineco: Done in a3f5107

config/300-awss3.yaml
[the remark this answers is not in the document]
antoineco: I ended up improving the current situation in 742dba9 by
* handling the `com.amazon.s3.testevent` case
* falling back to `com.amazon.sqs.message` when the received message can not be identified as originating from S3

The second case would indicate a flaw in the way customers use their SQS queue, but at least we try to handle things gracefully instead of discarding messages altogether.

While I was at it, I also removed the usage of `ceOverrides` in `adapter.go`, because it was unnecessary and made the origin of the "source" attribute unclear.

sameersbn:
@antoineco just noticing that this PR did not add the `AWSS3SOURCE_IMAGE` env variable in the controller.yaml

https://github.com/triggermesh/aws-event-sources/blob/main/config/500-controller.yaml#L58-L77

pkg/reconciler/awss3source/adapter.go
```diff
@@ -0,0 +1,100 @@
...
+const envMessageProcessor = "SQS_MESSAGE_PROCESSOR"
+
+const healthPortName = "health"
+
+// adapterConfig contains properties used to configure the source's adapter.
+// These are automatically populated by envconfig.
+type adapterConfig struct {
+	// Container image
+	// Reuses the adapter from the SQS source.
+	Image string `envconfig:"AWSSQSSOURCE_IMAGE" default:"gcr.io/triggermesh/awssqssource"`
```
sameersbn, line 44: @antoineco should this be "Image string `envconfig:"AWSS3SOURCE_IMAGE" default:"gcr.io/triggermesh/awss3source"`"?

antoineco: The S3 source consumes bucket notifications over a SQS queue (which we reconcile in the controller if it doesn't exist), therefore the S3 source simply reuses the SQS source's adapter.

PR #272, opened 2021-04-06
Disallow empty name/key in valueFromSecret refs

antoineco:
@JeffNeff noticed today that it was currently possible to set a `valueFromSecret` without name or key. This is not expected.

This PR marks those fields as required in OpenAPI schemas.


JeffNeff (approved):
lgtm

PR #289, opened 2021-05-21
Refactor Spec

JeffNeff:
Removed identifier & serviceType from the spec. We are now pulling the ResourceID from the provided ARN. 

I also slid in ARN Validation to the CRD and updated the Event type

An example of the new spec:
```
apiVersion: sources.triggermesh.io/v1alpha1
kind: AWSPerformanceInsightsSource
metadata:
    name: pi
spec:
    arn: arn:aws:rds:us-west-2:925906438773:db:database-2-instance-1
    pollingInterval: 1m
    metricQueries: 
      - os.cpuUtilization.idle.avg
      - os.general.numVCPUs.avg
      - os.network.rx.avg
      - os.network.tx.avg
      - os.network.rx.avg
      - os.fileSys.total.avg 
      - os.fileSys.used.avg 
      - os.memory.active.avg
      - os.memory.total.avg
      - os.tasks.blocked.avg
      - os.tasks.zombie.avg
    credentials:
        accessKeyID:
            valueFromSecret:
                key: access_key_id
                name: aws
        secretAccessKey:
            valueFromSecret:
                key: secret_access_key
                name: aws
    sink:
        ref:
            apiVersion: serving.knative.dev/v1
            kind: Service
            name: event-display
```

### DIFF
--- a/pkg/adapter/awsperformanceinsightssource/adapter.go
+++ b/pkg/adapter/awsperformanceinsightssource/adapter.go
@@ -28,6 +28,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/pi"
+	"github.com/aws/aws-sdk-go/service/rds"
 
 	pkgadapter "knative.dev/eventing/pkg/adapter/v2"
 	"knative.dev/pkg/logging"
@@ -35,6 +36,8 @@ import (
 	"github.com/triggermesh/aws-event-sources/pkg/adapter/common"
 	"github.com/triggermesh/aws-event-sources/pkg/apis/sources/v1alpha1"
 )
+
+const serviceType = "RDS"
 
 // envConfig is a set parameters sourced from the environment for the source's
 // adapter.
@@ -46,10 +49,6 @@ type envConfig struct {
 	PollingInterval string `envconfig:"POLLING_INTERVAL" required:"true"`
 
 	MetricQueries []string `envconfig:"METRIC_QUERIES" required:"true"`
-
-	Identifier string `envconfig:"IDENTIFIER" required:"true"`
-
-	ServiceType string `envconfig:"SERVICE_TYPE" required:"true" `
 }
 
 // adapter implements the source's adapter.
@@ -62,8 +61,7 @@ type adapter struct {
 	arn             arn.ARN
 	pollingInterval time.Duration
 	metricQueries   []*pi.MetricQuery
-	identifier      string
-	serviceType     string
+	resourceID      string
 }
 
 // event represents the structured event data to be sent as the payload of the Cloudevent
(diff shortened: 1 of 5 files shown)


pkg/adapter/awsperformanceinsightssource/adapter.go
```diff
@@ -62,8 +61,7 @@ type adapter struct {
 	arn             arn.ARN
 	pollingInterval time.Duration
 	metricQueries   []*pi.MetricQuery
-	identifier      string
-	serviceType     string
+	resourceID      string
```
antoineco, line 64: Are we sure the RDS id the same as the PI id?
I did a test last week and they were different. The PI id was "db-longrandomstring" whereas the RDS id was the name I gave my insurance?

pkg/adapter/awsperformanceinsightssource/adapter.go
```diff
@@ -102,6 +100,19 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClie
...
+	r := rds.New(cfg)
+	var resourceID string
+	dbi, err := r.DescribeDBInstances(&rds.DescribeDBInstancesInput{})
+	if err != nil {
+		logger.Panicf("Unable describe DB clusters: %v", zap.Error(err))
+	}
+
+	for _, instance := range dbi.DBInstances {
+		if *instance.DBInstanceArn == a.String() {
+			resourceID = *instance.DbiResourceId
```
antoineco, line 112: Ah ok, I see what you're doing here. Didn't even know you could correlate the 2 IDs like that 👌

antoineco (approved):
Now I understand the API :)
A few nits

JeffNeff:
```
nit: this reads as "Performance Insight Insight".
Maybe we could name the type ...pi.metric?
```
It reads 
`Performance Insights insight`. I think that is okay and is better aligned with docs/api. But I am okay either way if you have hard feelings on `pi.metric` we can go that way!

pkg/adapter/awsperformanceinsightssource/adapter.go
```diff
@@ -165,9 +175,8 @@ func (a *adapter) PollMetrics(priorTime time.Time, currentTime time.Time) {
 				}
 
 				event := cloudevents.NewEvent(cloudevents.VersionV1)
-				event.SetType(v1alpha1.AWSEventType(a.arn.Service, v1alpha1.AWSPerformanceInsightsGenericEventType))
+				event.SetType(v1alpha1.AWSPerformanceInsightsGenericEventType)
 				event.SetSource(*d.Key.Metric)
```
antoineco, line 179: Maybe the source could also be the ARN now, like in other AWS sources.
And you can use an extension for the metric (`ceer.SetExtension("pimetric", ...)`) in case people want to filter on that.

pkg/apis/sources/v1alpha1/awsperformanceinsights_lifecycle.go
```diff
@@ -53,7 +53,7 @@ func (s *AWSPerformanceInsightsSource) GetStatusManager() *EventSourceStatusMana
 
 // Supported event types
 const (
-	AWSPerformanceInsightsGenericEventType = "pi.insight"
+	AWSPerformanceInsightsGenericEventType = "com.amazon.rds.pi.insight"
```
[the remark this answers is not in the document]
antoineco: @JeffNeff is what you are sending called "an insight" or "a metric"?
If that's the former I'm good with the current version.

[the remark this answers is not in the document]
JeffNeff: you are right! its a metric.

PR #290, opened 2021-05-27
retry SNS reconciliation if InvalidParameter

odacremolbap:
`InvalidParameter` code might be [returned from AWS](https://docs.aws.amazon.com/sns/latest/api/API_Subscribe.html#API_Subscribe_Errors) as the result of an unsuccessful subscribe operation.

The parameter provided from the controller using the external URL for the source, chances that the parameter is invalid (wrong URL) are low. It looks like at subscription time AWS tests the endpoint and if it is not ready yet, the `InvalidParameter` will be returned, making the reconciliation fail.

For such cases, that error code should be treated as non permanent at the SNS source.

### DIFF
--- a/pkg/reconciler/awssnssource/subscription.go
+++ b/pkg/reconciler/awssnssource/subscription.go
@@ -91,7 +91,7 @@ func (r *Reconciler) ensureSubscribed(ctx context.Context) error {
 			//
 			// However "InvalidParameter" is returned if the subscription URL is not ready, which
 			// might be the case if the endpoint for the source is not exposed yet.
-			if awsErr := awserr.Error(nil); errors.As(err, &awsErr) && awsErr.Code() != "InvalidParameter" {
+			if awsErr := awserr.Error(nil); errors.As(err, &awsErr) && awsErr.Code() == sns.ErrCodeInvalidParameterException {
 				return event
 			}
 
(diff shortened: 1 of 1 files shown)


pkg/reconciler/awssnssource/subscription.go
```diff
@@ -82,11 +82,20 @@ func (r *Reconciler) ensureSubscribed(ctx context.Context) error {
...
+
+			// Most documented API errors require some user intervention and
 			// are not to be retried.
 			// https://docs.aws.amazon.com/sns/latest/api/API_Subscribe.html#API_Subscribe_Errors
-			status.MarkNotSubscribed(v1alpha1.AWSSNSReasonRejected, "Subscription request rejected")
-			return controller.NewPermanentError(subscribeErrorEvent(url, topicARN, err))
+			//
+			// However "InvalidParameter" is returned if the subscription URL is not ready, which
+			// might be the case if the endpoint for the source is not exposed yet.
+			if awsErr := awserr.Error(nil); errors.As(err, &awsErr) && awsErr.Code() != "InvalidParameter" {
```
tzununbekov, line 94: `sns.ErrCodeInvalidParameterException` can be used as the error code :slightly_smiling_face:

tzununbekov (approved):

pkg/reconciler/awssnssource/subscription.go
odacremolbap: thanks a lot!
fixing

odacremolbap: self review: also it should not be `!=` but `==` when the non permanent error is returned

odacremolbap: https://github.com/triggermesh/aws-event-sources/pull/290/commits/0d71f6a484f1cbea0e6f41f774b5edd95054fd5f

PR #294, opened 2021-06-09
ci: attach manifests YAMLS to releases

### LINKED ISSUE #274: Idea: allow fan-out over SNS topic
Currently, there is no possibility for _multiple_ subscribers to create a notification configuration in the _same_ bucket. AWS rejects the request with the error `Configurations on the same bucket cannot share a common event type.`.

Even if we enforced a mechanism such as allowing multiple subscribers only if they define the exact same list of event types, we would be limited by the feature of the SQS queue, which doesn't allow fanning messages out to multiple consumers.

We could, however, enable an alternative mode where a user could pass a `destinationARN` pointing at a SNS topic (which must exist) instead of a SQS queue, and the source would then create and subscribe one SQS queue to that topic, for each AWSS3Source instance which uses that mode.

Ref. https://docs.aws.amazon.com/sns/latest/dg/sns-sqs-as-subscriber.html

sameersbn:
Closes #274

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,10 @@
 version: 2.1
 
 orbs:
-  go: circleci/go@1
   gcp-cli: circleci/gcp-cli@1
+  go: circleci/go@1
   helm: circleci/helm@1
+  k8s: circleci/kubernetes@0
 
 jobs:
   checkout:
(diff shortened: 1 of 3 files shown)


sebgoa (approved):
lgtm

PR #302, opened 2021-07-20
Kustomization base added

tzununbekov:
As a part of deployment optimization described in [here](https://github.com/triggermesh/config/issues/516), this repository should act as a remote base for Kustomization. `kustomization.yaml` with deployment resources added to the repository root to keep `config` directory ko apply-able.


config/kustomization.yaml
```diff
@@ -1,19 +0,0 @@
...
-- 300-awscognitouserpool.yaml
-- 300-awsdynamodb.yaml
-- 300-awskinesis.yaml
-- 300-awsperformanceinsights.yaml
-- 300-awss3.yaml
-- 300-awssns.yaml
-- 300-awssqs.yaml
-- 500-controller.yaml
-
-namespace: triggermesh
```
sameersbn, line 19: recently we added this file so that the CI could generate the manifest for attaching to releases (ref: https://github.com/triggermesh/aws-event-sources/pull/294). This requires that the `namespace.yaml` manifest and the `namespace` configuration to exist so that the yaml is rendered properly.

This PR would break it

sameersbn: maybe we could add another `kustomize_release.yaml` for this purpose?

tzununbekov: oh, didn't know that this kustomization.yaml was there for a purpose... I'd think about other options to generate release artifacts to keep `config` directory ready for kubectl/ko `apply` command.

sameersbn: are you aware of an alternate way of generating it, else we'd have to comment out these lines https://github.com/triggermesh/aws-event-sources/blob/master/.circleci/config.yml#L129-L131 in the interim

tzununbekov: one of the options, as you said, is to add second kustomization yaml for release. Don't really like it though, need to think a bit about it

sameersbn (approved):
approving.. have made a note in https://github.com/triggermesh/misc/issues/258

PR #327, opened 2021-09-11
Merge DynamoDB event types into single type

### LINKED ISSUE #326: Merge insert/modify/delete event types into a single event type
Declaring multiple event types for DynamoDB increases the complexity of CloudEvents unnecessarily because all 3 events use a single schema (unlike sources such as GitHub, GitLab, CodeCommit, ...).

https://github.com/triggermesh/aws-event-sources/blob/13d3cd242c94c32ff7539c541fa83ed5cd0e9416/config/300-awsdynamodb.yaml#L27-L29

Let's merge them into a single event type (e.g. [`com.amazon.dynamodb.stream_record`](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_streams_GetRecords.html)) and set the name of the operation as a `operation` CloudEvents extension attribute instead.

antoineco:
Closes #326 

All DynamoDB events have now the same CloudEvents type, `com.amazon.dynamodb.stream_record`, regardless of the operation that was performed (insert, modify, remove).
The operation (type of data modification) is now provided as a CloudEvents extension attribute: `dynamodboperation`.

The rationale is explained in the linked issue.

---

Examples of Knative Trigger filters relevant for filtering events from this source after this change:

```yaml
# ALL events from DynamoDB
# (required 3 separate Triggers before this change)
filter:
  attributes:
    type: com.amazon.dynamodb.stream_record
```

```yaml
# Only table INSERTs
filter:
  attributes:
    type: com.amazon.dynamodb.stream_record
    dynamodboperation: INSERT
```

```yaml
# Events originating from a certain DynamoDB table
# (already possible before this PR, but filtering on 'type' required multiple Triggers)
filter:
  attributes:
    type: com.amazon.dynamodb.stream_record
    source: arn:aws:dynamodb:us-west-2:123456789012:table/triggermeshtest
```


odacremolbap (approved):
I'm 100% on merging this, but also commenting; while I like this change, and personally think we tend to underuse extended attributes, we will need to take into account that part of the semantics now live in a header extension, and that the future event registry will need to cover them.

Docs might need to add a reference to this new attribute.

antoineco:
Absolutely. I'm going to leave this open and think about it for a bit, maybe until we discuss the consequences.

The next question would be whether we want to do the same thing for [S3 events](https://github.com/triggermesh/aws-event-sources/pull/267), which also use a single schema regardless of the event type.

PR #331, opened 2021-09-20
sqs: Add option for configuring visibility timeout

### LINKED ISSUE #329: Allow customization of SQSSource visibility timeout
Currently when reading messages from SQS the visibility timeout is [set at a static 30 seconds](https://github.com/triggermesh/aws-event-sources/blob/main/pkg/adapter/awssqssource/receive.go#L43). However, depending on the type of sink configured this may be insufficient.

I have a long-running service as the sink in my configuration. The goal is to maintain SQS as the source of truth for whether or not a message has been processed, not deleting the message from the queue until the service returns successfully. This works today with the exception of the visibility timeout - it ends up in an infinite retry loop because the response from the sink takes >30 seconds even though each individual request is successful.

Ideally the visibility timeout should be configurable from the spec.

antoineco:
Adds a new option `receiveOptions.visibilityTimeout` which controls the [visibility timeout](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html) set on received messages:

```yaml
apiVersion: sources.triggermesh.io/v1alpha1
kind: AWSSQSSource
metadata:
  name: sample
spec:
  # ...
  receiveOptions:
    visibilityTimeout: <duration string; e.g. "1m30s">
```

The previously hardcoded value of `30s` has been removed and, when the user doesn't set a value at all, the overall value configured for the queue is used.

Closes #329

cc. @noahw3


JeffNeff (approved):
lgtm

PR #333, opened 2021-09-21
pi: Rename metricQueries to metrics

### LINKED ISSUE #284: MetricQuery supports neither GroupBy nor Filter
The current spec contains a single `metricQuery` field which semantics is confusing because it doesn't translate to a full metric query, but rather to the specific `Metric` field of a metric query:

https://github.com/triggermesh/aws-event-sources/blob/56307ae71d80c3d2c9db2931d7e0f3e0aabfb2fa/pkg/adapter/awsperformanceinsightssource/adapter.go#L93

API doc:
https://docs.aws.amazon.com/performance-insights/latest/APIReference/API_GetResourceMetrics.html

If we ever want to let users express full metric queries, we will need to switch to a more structured spec, such as (expressed as OpenAPI schema):

```yaml
metricQuery:
  type: object
  properties:
    metric:
      type: string
    filter:
      type: string
    groupBy:
      type: object
      properties:
        group:
          type: string
        dimensions:
          type: array
          items:
            type: string
        limit:
          type: integer
      required:
      - group
  required:
  - metric
```

The Go structs used within the AWS SDK are documented here:
- [pi.MetricQuery](https://docs.aws.amazon.com/sdk-for-go/api/service/pi/#MetricQuery)
- [pi.DimensionGroup](https://docs.aws.amazon.com/sdk-for-go/api/service/pi/#DimensionGroup)

cc @JeffNeff FYI

antoineco:
Closes #284

tl;dr: _What we call a "metrics_query" is not the same things as the "metric_query" defined by Amazon._

As discussed in the referenced issue, this source only allows users to request a list of **metrics**, without the `GroupBy` and `Filter` attributes available inside Amazon's actual [`MetricQueries`](https://docs.aws.amazon.com/performance-insights/latest/APIReference/API_MetricQuery.html).

In a effort to keep the vocabulary of our APIs as accurate as possible, I suggest we rename

```yaml
metric_queries: []
```

to

```yaml
metrics: []
```

### DIFF
--- a/config/300-awsperformanceinsights.yaml
+++ b/config/300-awsperformanceinsights.yaml
@@ -61,10 +61,10 @@ spec:
                 description: Duration which defines how often metrics should be pulled from Amazon Performance Insights.
                   Expressed as a duration string, which format is documented at https://pkg.go.dev/time#ParseDuration.
                 type: string
-              metricQueries:
-                description: List of queries that determine what metrics will be sourced from Amazon Performance
-                  Insights. Each item represents the 'metric' attribute of a MetricQuery. For more information, please
-                  refer to the Performance Insights API reference at
+              metrics:
+                description: List of metrics to retrieve from Amazon Performance Insights. Each item represents the
+                  'metric' attribute of a MetricQuery. For more information, please refer to the Performance Insights API
+                  reference at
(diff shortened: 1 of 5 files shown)


JeffNeff (approved):
lgtm

PR #334, opened 2021-09-21
sqs: Include all message attributes in event data

### LINKED ISSUE #330: Message attributes aren't propagated to CloudEvents
While receiving messages, we don't request MessageAttributeNames, therefore part of the information contained in the original message might be lost.

antoineco:
SQS message attributes were not fetched at all until now, therefore we never included them into our events' data.
This PR ensures all message attributes are included.

In a future improvement, we may offer users the option to select what attributes they are interested in, instead of defaulting to "All".

Fixes #330


JeffNeff (approved):
nice catch. lgtm

PR #338, opened 2021-09-23
Remove empty ClusterRoles

antoineco:
All these `ClusterRoles`/`ClusterRoleBindings` were originally added for consistency, but `AWSSNSSource` is the only source which adapter interacts with the Kubernetes API, so I don't see any reason to keep creating these empty objects. It just creates noise.


JeffNeff (approved):
lgtm

